PR #12, opened 2025-03-20
Refactor line search

benedict-96:
Wrote documentation for more or less the entire package. 
Major changes concern:
1. *backtracking line search*: this has been completely reworked. This was previously also not tested for as all relevant tests were commented out.
2. `bracketing` was only defined for minima before (`bracket_minimum`). This pr implements `bracket_root`.
3. Implemented `BierlaireQuadratic` found in https://transp-or.epfl.ch/books/optimization/html/OptimizationPrinciplesAlgorithms2018.pdf as an alternative to `Quadratic`.
4. Bisections: minimum version.

Open questions:
- [x] Should we really distinguish between `UnivariateObjective`s and `MultivariateObjective`s? (or just store `Vector`s of length one for the former).
- [x]  Do we really need `value`, `value!`, `value!!` etc.?
- [ ]  Use of in-place and out-of-place functions doesn't seem to be consistent. See e.g. `abstract_newton_solver.jl`.
- [ ]  When computing `residual!(::NonlinearSolverStatus)` we are doing `rxₐ / rxₐ`. This doesn't seem to make much sense.
- [ ]  `state(opt)` gives an object of type `OptimizationAlgorithm` (instead of `OptimizationState`). We have however `NewtonOptimizerState <: OptimizationAlgorithm`.
- [ ]  Should `JacobianFunction` also store `F` or only `DF!`?
- [x]  In `linesearch_objective` the function `d` computes the negative of the gradient. Is this a good choice? 
- [x] Should we use a `TemporaryUnivariateObjective` that only depends on $\alpha$ when doing backtracking (as is done now) or should we work with the `MultivariateObjective` directly?
- [ ] The keyword `algorithm` in `Optimizer` has to be a `NewtonMethod`. It could also be a different method (i.e. non-Newton style).
- [ ] Should we have consistent use of `Optimizer` and `Optimization`? In e.g. `OptimizerState` and `OptimizationAlgorithm`.
- [ ] We should probably combine the two `linesearch_objective` methods.
- [ ] `NewtonOptimizerCache` and `MultivariateObjective` both store the value of the gradient.
- [x] the roles of `update!` and `initialize!` should be clarified. Should initialize really set some variables to `NaN`s? At the moment we first call `initialize!` and then `update!` to really initialize ...
- [ ] Should we store an array in e.g. `NewtonOptimizerCache` or an instance of `Gradient`?
- [ ] Why does `Static` also save α (similar to `StaticState`)? This is not the same for the other line search algorithms (i.e. seems to be an inconsistency).
- [ ] Do we need the extra `struct` `Linesearch`? Are `LinesearchMethod`s and `LinesearchState`s not enough? (Because every `LinesearchState` corresponds to a `LinesearchMethod`. `Linesearch` also stores an instance of `Options`; maybe this is needed somewhere?); see e.g. `newton_solver.jl`!
- [x] Should we have a dependency on `NLsolve`? (see `nlsolve_newton.jl`).
- [ ] We have a method `Newton <: NewtonMethod`. Maybe this should have a different name? Perhaps `AD <: NewtonMethod`? 
- [ ] Should we get rid of routines `gradient_ad!` and `gradient_fd!`? This doesn't add much more functionality but introduces additional terminology.
- [ ] Are bisections also a form ob "bracketing" or does "bracketing" refer to finding a bracket? (https://www.uoanbar.edu.iq/eStoreImages/Bank/3027.pdf). Should we use the `abstract type` `BracketingMethod` here?
- [ ] When calling `NewtonSolver` we should probably add the keyword `linear_solver`, to be passed to `LinearSolver`.
- [ ] Should `derivative` in `SimpleSolvers` extend `ForwardDiff.derivative`?
- [ ] We should use a consistent notation for *the function interpreted as the objective*, now we use `F`, `f`, `F!`, ...
- [x] Calling `linesearch_objective` for solvers returns the norm of the objective, whereas `linesearch_objective` applied to optimizers just returns the value of the objective (I added a comment regarding this in the corresponding docstring)
- [ ] Only the "root-finding" version of `Bisections`s was implemented. I extended this to include a "minimum-finding" version. Do we only need this one?
- [ ] Should we derive a `linesearch_objective` only from another `AbstractObjective`? This is currently done that way for optimizers but not for solvers.
- [ ] Should we use the `CurvatureCondition` only for `Backtracking` but also for `Quadratic` and `BierlaireQuadratic`?
- [ ] When we call `update!` on `NewtonOptimizerCache` should we also update the Hessian? (this is not done at the moment).
- [ ] Should we also have a routine `update!(cache, x, grad)` or only `update!(cache, x, grad, hes)`?
- [ ] Should we call `solver_step!` on only the `NewtonSolver` (it should contain all necessary arguments if we add the objective)?
- [ ] We will have to distinguish between `MultivariateObjective`s for solvers and optimizers as the former returns a vector and the latter a scalar. For now a `MultivariateObjective` only stores a `Gradient`, in the solver case it should store a `Jacobian`.

General questions:
- [ ] How should we deal with Hessians near a saddle point or local extremum?
- [ ] How and where should we check whether we already reached a minimum when performing line search. The functor of `QuadraticBierlaireState` checks whether the derivative of the objective *is small*.


benedict-96:
I marked already discussed points as completed tasks:
1. yes, we should distinguish between `UnivariateObjective`s and `MultivariateObjective`s,
2. yes, we should have `value`, `value!`, `value!!` etc.
3. `initialize!` first allocates `NaN`s and the *numerical initialization* has to be done by calling `update!`. This will be clarified in the docs and the doc strings.

Issue #33, opened 2025-05-19
Do we need `initialize!` routines? `clear!` should be enough.


benedict-96:
The function `initialize!` always takes input arguments of the form `initialize!(object, x)`, but `object` seemingly always stores an `x`. Should we get rid of the `initialize!` function and instead just implement `clear!` routines which only take one input argument?

benedict-96:
I implemented `initialize!` routines for (I think) every object now in https://github.com/JuliaGNI/SimpleSolvers.jl/pull/35. Closing https://github.com/JuliaGNI/SimpleSolvers.jl/pull/35 should solve this issue.

Issue #31, opened 2025-05-12
Redundancy in `NonlinearSolverStatus`, `NewtonSolverCache` and `MultivariateObjective`


benedict-96:
`NonlinearSolverStatus`, `NewtonSolverCache` and `MultivariateObjective` are all storing the current iterate `x` as well as `x₁`/`x̄` (`NewtonSolverCache`/`NonlinearSolverStatus`).

Issue #29, opened 2025-05-11
Should we call `solver_step!` on only the NewtonSolver (i.e. with a single argument)?


benedict-96:
It would perhaps be most elegant if we wrote:
```julia
solver_step!(solver)
```
An instance `solver` should contain all necessary arguments if we add the `objective` to it.

Issue #28, opened 2025-05-11
Difference between `MultivariateObjective` for *solver* and *optimizer*.


benedict-96:
We will have to distinguish between `MultivariateObjective`s for solvers and optimizers as the former returns a vector and the latter a scalar. For now a `MultivariateObjective` only stores a `Gradient`, in the solver case it should store a `Jacobian`.

PR #35, opened 2025-05-21
Refactor `NonlinearSolver`s

benedict-96:
Refactored `NonlinearSolver`s (especially `NewtonSolver`):
- combined `NewtonSolver` and `QuasiNewtonSolver` under one `struct`,
- defined `LinearSystem` and `NonlinearSystem`
- refactored `LinearSolver`s and associated routines `factorize!`, `solve` and `solve!` to make their use more intuitive.

Regarding linesearches:
- implemented a new `Quadratic2` line search that works better than the old one (called `Quadratic`). We may should consider getting rid of this old version.

Miscellaneous:
- consistently implemented the `initialize!`-`update!` interface, i.e. the constructor always calls `initialize!` first, not matter the object we are treating. In order to start *solving a system* we always have to call `update!`.

Related issues:
- closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/21
- closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/24 (the keyword is `linearsolver`; also see `LinearSolver`)
- closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/28 - this is being solved with the introduction of `NonlinearSystem`s.
- potentially closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/29 - at the moment `solver_step!` is still called with `x` as an additional input argument
- closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/31 - we get rid of some of this redundancy by not storing `J` anymore in the `NewtonSolverCache` for example.
- closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/33 - `initialize!` has been implemented for most objects; sometimes by internally calling `clear!` (see e.g. `initialize!(::OptimizerResult, AbstractVector)`).
- closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/25 - this has been solved by including `Hessian` in the `Optimizer` `struct` and removing it from `NewtonOptimizerState`.
- closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/15 - we now compute the relative residual for the function value by comparing with $f_0$ (was taken from PETSc manual??).


michakraus:
- [ ] Should an `AbstractObjective` be a subtype of `AbstractProblem` ?

src/base/options.jl
```diff
@@ -162,22 +166,28 @@ function Options(;
                         verbosity)
 end
 
+convert_float64_to_float32_if_necessary(a::Number) = convert_float64_to_float32_if_necessary(Val(a))
```
michakraus, line 169: This functions seems non-obvious. Needs to be discussed before merge.

src/base/systems.jl
```diff
@@ -0,0 +1,343 @@
...
+
+"""
+    clear!(ls)
+
+Write `NaN`s into `Matrix(ls)` and `Vector(ls)`.
+"""
+function clear!(ls::LinearSystem)
+    Matrix(ls) .= alloc_j(Matrix(ls)[1, :], Vector(ls))
+    Vector(ls) .= alloc_x(Vector(ls))
+    nothing
```
michakraus, line 106: These seems inconsistent. Most similar functions (`update!`, etc.) return the object.

src/base/systems.jl
```diff
@@ -0,0 +1,343 @@
...
+    nls.f_calls = 0
+    f_argument(nls) .= alloc_x(f_argument(nls))
+    value(nls) .= alloc_f(f_argument(nls))
+    nothing
+end
+
+function _clear_j!(nls::NonlinearSystem)
+    nls.j_calls = 0
+    j_argument(nls) .= alloc_x(j_argument(nls))
+    jacobian(nls) .= alloc_j(j_argument(nls), value(nls))
```
michakraus, line 264: Same `alloc_x` and `alloc_j` issue as mentioned above.

src/linear/lu_solver.jl
```diff
@@ -78,67 +162,80 @@ lu.A
...
+        cache(lsolver).perms[i] = i
     end
 
-    @inbounds for k ∈ 1:lu.n
-        # find index max
-        kp = pivot ? find_maximum_value(lu.A[:, k], k) : k
+    n = size(cache(lsolver).A, 1)
+
+    @inbounds for k ∈ axes(cache(lsolver).A, 1)
+        kp = method(lsolver).pivot ? find_maximum_value(cache(lsolver).A[:, k], k) : k
```
michakraus, line 193: `cache(lsolver).A[:, k]` creates a copy. One should use views here.

benedict-96:
> * [ ]  Should an `AbstractObjective` be a subtype of `AbstractProblem` ?

I'm not 100% sure, but we are also subtyping `NonlinearSystem` and `LinearSystem` from `AbstractProblem`.
In `GeometricBase` there is also no definition of what should be an `AbstractProblem`. We could perhaps define it as everything that has a solution?

src/base/objectives.jl
```diff
@@ -368,41 +396,62 @@ end
 
 function _clear_f!(obj::MultivariateObjective{T, Tx, TF, TG, Tf}) where {T, Tx, TF, TG, Tf <: Number}
     obj.f_calls = 0
-    obj.f = alloc_f(obj.x_f)
-    obj.x_f .= alloc_x(obj.x_f)
+    obj.f = alloc_f(f_argument(obj))
+    f_argument(obj) .= alloc_x(f_argument(obj))
```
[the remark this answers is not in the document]
benedict-96: Yes, I think that makes much more sense!
We should perhaps consider getting rid of all the `alloc_`... functions or at least only have one of them, not `alloc_x`, `alloc_f`, `alloc_g` ...
It makes more sense, as you write, to have `f_argument(obj) .= zero(Tx)` or `f_argument(obj) .= Tx(NaN)`.

src/base/options.jl
benedict-96: It's purpose is to convert machine precision "if necessary". I.e. if we have `eps(Float64)` then it should be converted to `eps(Float32)`. The specific implementation of this functionality is however questionable ...

src/base/systems.jl
benedict-96: Fixed in https://github.com/JuliaGNI/SimpleSolvers.jl/pull/35/commits/e639e024503b80da4b9137c85c195d452512b49b.

src/linear/lu_solver.jl
benedict-96: Solved with https://github.com/JuliaGNI/SimpleSolvers.jl/pull/35/commits/c00c4445d1d9ba4c63e683e68589e2e4cb2a621a.

src/base/options.jl
benedict-96: This comment was hopefully solved with [fd7b776](https://github.com/JuliaGNI/SimpleSolvers.jl/pull/35/commits/fd7b776a83acfe8a77a0f98c59cad7d1be9e5bf1).

benedict-96:
> > * [ ]   Should an `AbstractObjective` be a subtype of `AbstractProblem` ?
> 
> I'm not 100% sure, but we are also subtyping `NonlinearSystem` and `LinearSystem` from `AbstractProblem`. In `GeometricBase` there is also no definition of what should be an `AbstractProblem`. We could perhaps define it as everything that has a solution?

`AbstractObjective` is now NOT subtyped from `AbstractProblem`. This was changed in https://github.com/JuliaGNI/SimpleSolvers.jl/pull/35/commits/225095cb46501b8f4bb173083355ffb436d1b454.

PR #68, opened 2025-11-13
`MultivariateOptimizerProblem` -> `OptimizerProblem`

benedict-96:
This pr further removes `UnivariateOptimizerProblem` and renames the `TemporaryUnivariateProblem` to `LinesearchProblem`.


benedict-96:
> [!NOTE] 
> In the commit message in https://github.com/JuliaGNI/SimpleSolvers.jl/pull/68/commits/905407bc2cfa35b82aa8105e193a5f78e302eebc we should have *problem* instead of *objective*.

Issue #90, opened 2025-12-10
Move functionality from `SolverProblem` to `SolverCache`


benedict-96:
And fix associated problems with `value!` and `value!!` etc.

PR #96, opened 2025-12-10
Removed arrays from `NonlinearSolverProblems`

benedict-96:
Removed arrays from `NonlinearSolverProblems` (including associated functionality) and moved relevant bits to cache.

Closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/90


src/nonlinear/newton_solver.jl
michakraus:
```suggestion
        value!(cache(s).y, nonlinearproblem(s), x, params)
        if any(isnan, cache(s).y)
```

Issue #41, opened 2025-06-01
Should we remove the old `Quadratic` line search?


benedict-96:
In https://github.com/JuliaGNI/SimpleSolvers.jl/pull/35 we implemented a new line search called `Quadratic2` that works better in practice than the old `Quadratic` and is more intuitive. We should probably get rid of the old one.

PR #99, opened 2025-12-12
Moved old quadratic linesearch that's not used anymore to `obsolete`

benedict-96:
Moved old quadratic linesearch that's not used anymore to directory "obsolete." The one we use is in "custom_quadratic.jl" and called "Quadratic2" (this name should probably be changed).

Closes https://github.com/JuliaGNI/SimpleSolvers.jl/issues/41.


src/linesearch/bisection.jl
```diff
@@ -33,7 +33,7 @@ and then repeat:
 ```
 So the algorithm checks in each step where the sign change occurred and moves the ``x_0`` or ``x_1`` accordingly. The loop is terminated (and errors) if `config.max_iterations` is reached (by default""" * """$(MAX_ITERATIONS) and the [`Options`](@ref) struct).
 
-!!! warning
+!!! warninging
```
michakraus:
```suggestion
!!! warning
```

PR #142, opened 2026-02-03
Implementing Powell's Dogleg

benedict-96:
This pr implements the [dogleg method](https://en.wikipedia.org/wiki/Powell%27s_dog_leg_method#cite_note-powell1970-1).

We should beforehand clarify a few points:
- [x] should we implement a separate cache for the dogleg? It requires *two directional vectors*? On [wikipedia](https://en.wikipedia.org/wiki/Powell%27s_dog_leg_method#cite_note-powell1970-1) they are called ${\delta_{gn}}$ and $\delta_{sd}$. At the moment the cache only stores one directional vector (which we call $\Delta x$.
- [ ] should linesearch be applied to the *final direction*?
- [x] how is $s$ chosen in $t\delta_{sd} + s(\delta_{gn} - t\delta_{sd})$?

### DIFF
--- a/src/nonlinear/dogleg_solver.jl
+++ b/src/nonlinear/dogleg_solver.jl
@@ -40,15 +40,19 @@ function directions!(s::DogLegSolver{T}, x::AbstractVector{T}, params) where {T}
 
     # the steepest descent direction
 
-    direction₁(cache(s)) .= rhs(linearproblem(s))
-    direction₁(cache(s)) .*= L2norm(rhs(linearproblem(s)))
-    direction₁(cache(s)) ./= (rhs(linearproblem(s)) ⋅ (jacobianmatrix(s) * rhs(linearproblem(s))))
+    mul!(direction₁(cache(s)), transpose(jacobianmatrix(s)), rhs(linearproblem(s)))
+    fac₁ = L2norm(direction₁(cache(s)))
+    mul!(cache(s).y₂, jacobianmatrix(s), direction₁(cache(s)))
+    mul!(cache(s).y₃, transpose(jacobianmatrix(s)), cache(s).y₂)
+    fac₂ = direction₁(cache(s)) ⋅ cache(s).y₃
+    direction₁(cache(s)) .*= fac₁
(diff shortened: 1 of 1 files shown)


benedict-96:
I implemented the Gauss-Newton direction in https://github.com/JuliaGNI/SimpleSolvers.jl/pull/142/changes/375db9b8e154ec77fb10d96f9ca83dd76f936678.

Two adjustments will probably have to be made:
1. We should use two different `LinearSystem`s to compute `direction₁` and `direction₂` (the same one is used twice at the moment).
2. After the changes in https://github.com/JuliaGNI/SimpleSolvers.jl/pull/142/changes/375db9b8e154ec77fb10d96f9ca83dd76f936678 we call `transpose(jacobianmatrix(s))` twice. The matrix transpose should probably also be stored in the cache to reduce allocations and redundancies.

benedict-96:
In https://github.com/JuliaGNI/SimpleSolvers.jl/pull/142/changes/5d95b9373e4173b5c50323b4334b71fc08148c4c I tackled two points raised in the initial message of this pull request:
1. The cache is storing *three direction vectors* now: (i) the Newton direction, the (ii) Gauss-Newton direction and (iii) a combination of the two.
2. The combination of the two is computed via $s$: https://github.com/JuliaGNI/SimpleSolvers.jl/blob/5d95b9373e4173b5c50323b4334b71fc08148c4c/src/nonlinear/dogleg_solver.jl#L86 This equation was taking by solving for $||d|| = \Delta$.

src/nonlinear/dogleg_cache.jl
```diff
@@ -12,12 +12,14 @@ struct DogLegCache{T,AT<:AbstractVector{T},JT<:AbstractMatrix{T}} <: AbstractNon
 
     rhs::AT
     y::AT
+    y₂::AT
+    y₃::AT
```
benedict-96: $y_3 = J_k^Ty_2$. The vectors $y_2$ and $y_3$ are used to compute the factor for the Cauchy point (solver version) without unnecessary allocations.

PR #144, opened 2026-02-05
Revise and unify line search interface

michakraus:


benedict-96:
After getting rid of the *vector-valued version* of the `SufficientDecreaseCondition` and the `CurvatureCondition`, we can also remove the $p_k$ in the `SufficientDecreaseCondition` as this is set to 1 anyway:
https://github.com/JuliaGNI/SimpleSolvers.jl/blob/453ded551ae67f91937c7cf8027f5702e9f1bee3/src/linesearch/backtracking.jl#L161

> [!NOTE]
> It was not set to 1 in the vector-valued case, but was instead the directional vector[^1].

[^1]: The directional vector gets converted to unity when using the `LinesearchProblem`.

benedict-96:
Also regarding the previous point, we can probably remove $\alpha_0$ as this is going to be zero anyway.

benedict-96:
Some docstrings still have to be updated, e.g. for `Linesearch`:
https://github.com/JuliaGNI/SimpleSolvers.jl/blob/7351d2ebda5d405c8372fdb7b0913521a4ee09dc/src/linesearch/linesearch.jl#L21-L41

benedict-96:
Made docs run through with https://github.com/JuliaGNI/SimpleSolvers.jl/pull/144/changes/0b19088f59a698ae95ae31478422799e206378c2.
The contents of the docs, as well as many docstrings, need to be investigated though (see e.g. comment above).

PR #160, opened 2026-06-10
Syntax fix: callable type instead of tuple assignment

mlechu:
Lowering has a bug where an assignment to a `where`-wrapped tuple is treated like an anonymous function, and it's a very easy typo to make when trying to add a method to a callable type.  I found this syntax in two packages, and both seemed to be typos, so this PR corrects the syntax before I make it a proper error.  Issue reference: https://github.com/JuliaLang/julia/issues/62051


michakraus:
Thanks a lot!

PR #161, opened 2026-07-10
Bug sweep and hardening from the 2026-07 code review (v0.9.0)

michakraus:
Systematic bug sweep and hardening of SimpleSolvers, driven by a full code review (2026-07-09) that found **8 classes of runtime-verified bugs in exported paths**, a series of source-verified algorithmic defects, and a set of correctness risks that could make solvers **silently return wrong answers**. The work was executed as seven phases, each leaving the package in a working, testable state, with a failing-before / passing-after regression test for every behavioral fix.

**Version: 0.8.4 → 0.9.0** (pre-1.0 breaking ⇒ minor bump). See `CHANGELOG.md` for the consolidated user-facing entry.

## Headline correctness fixes

These two could silently corrupt results on the *main* solver paths:

1. **Backtracking line-search stall.** The shrink loop enforced both Wolfe conditions, but shrinking α makes the curvature condition *harder* to satisfy — for `f(α) = (α−100)²` from α = 1 it ran 1000 iterations and silently returned **α ≈ 9.3e-302**. The loop now terminates on sufficient decrease alone (Nocedal & Wright Alg. 3.1); curvature is an opt-in post-hoc check; iteration exhaustion warns and returns the last valid step. A proper **strong-Wolfe (bracket + zoom) line search** was added as the new exported `StrongWolfe` for genuine curvature enforcement.
2. **Stagnation reported as convergence.** The default convergence criteria were successive-change based with `f_abstol = 0`, so any stalled step (e.g. the denormal α above) yielded `rxₛ = rfₛ = 0` → "converged" at an arbitrarily large residual. The successive-change criteria are now **gated by the absolute residual** (`g_restol = √eps`). Notably, the repo's own "Newton fails on the Powell problem" test turned out to be this bug — with the fix, Newton *converges* there and the test assertion was flipped.

## Dead / broken exported API (removed or repaired)

* Removed 11 exported-but-undefined or dead-on-arrival names: `LUSolverLAPACK` (stale ccall wrapper, `MethodError` on construction), `BracketingMethod`, `IterativeMethod`, `algorithm`, `result`, `state`, `status`, `minimizer`, `gradient`, `hessian`, `print_hessian`.
* Repaired broken entry points: the generic `Jacobian{T}(F, nx, ny; mode)` backend selector, the 5-arg `solve` for line searches, `bracket_root(prob, params, x)`, single-`x` `bisection`, and the `LinearProblem` dimension assert (`length(y) == size(A, 1)` — non-square systems now construct).
* Replaced the ambiguous `Base.convert` catch-all (which broke `convert(Any, Static())`) with an internal `change_precision`.

## Algorithmic fixes

* **DogLeg solver**: trust-radius termination no longer verbosity-gated; correct model decrease (2FᵀJd, was dᵀJF — wrong for non-symmetric Jacobians); stationary-point guard (starting at the exact root converges instead of NaN → throw); unbounded per-shrink recursion (Jacobian + factorization per level) replaced by a full **ρ-based trust-region update** (N&W Alg. 4.1) with the radius carried across steps; NaN trial merits are rejected by shrinking the radius (the old independent d₁/d₂ rescaling destroyed the dogleg invariant ‖d₁‖ ≤ ‖d₂‖).
* **Finite differences**: non-square Jacobians (row loop iterated input instead of output indices); step `ϵ|xⱼ| + ϵ` (was exactly zero at `xⱼ = −1`); precision-aware default step `8√eps(T)` (Float32 FD used a ~1 ulp step → garbage).
* **PicardSolver**: `d = −F` is not a descent direction for the ‖F‖² merit, so routing it through a Wolfe line search was unsound (and, combined with the stagnation bug, produced false convergence). It is now a **residual-safeguarded fixed-point iteration** (halve α until the residual is non-increasing) and rejects a `linesearch` keyword.
* **LU solver**: singular systems throw `SingularException` instead of silently returning NaN/Inf (`info` reset per factorization and checked in `ldiv!`); aliased `ldiv!(x, ls, x)` handled; integer/rational matrices promote (`LinearSolver(LU(), [1 2; 3 4])` works, as the docstring always claimed).
* **Newton refactorization**: `mod(iteration−1, refactorize)` skipped factorization on a fresh state (`mod(−1, r) = r−1`), running `ldiv!` against a NaN-initialized cache.
* **Bracketing**: `triple_point_finder`'s recursive retry result was not returned, and its loop init produced a degenerate triple; `bracket_minimum`'s early exit bracketed a *maximum* under the minimum criterion.
* **Quadratic line searches**: accidental squaring of all `Quadratic(T, method)` defaults (ε² ≈ 3e-30 was below machine eps); interpolation denominators guarded with bisection fallback and bracket containment; iteration off-by-one.

## Type stability and API hygiene

* Removed phantom type parameters (`AT` from `NonlinearSolver`, eltype `T` from `NonlinearProblem`) and the dead `NonlinearPreconditioner`.
* LU cache type chosen by **dispatch** (StaticMatrix → `MMatrix`, otherwise `Matrix`) instead of a runtime size threshold — `LinearSolverCache(LU(), A)` is now inferable. Behavior change: the default `LU()` no longer promotes small plain matrices to `MMatrix`.
* `CurvatureCondition` mode as `Val` positional (was a runtime `Symbol` keyword); `Options` keywords `::Real` (accepts `x_abstol = 0`, rationals); `Newton{true}(refactorize)` constructable; assorted signature relaxations (`GradientFunction`, `NonlinearProblem` inner constructor, `Bisection(::Type{T})`).
* Removed error-swallowing fallbacks (`initialize!(x...)`, 1-arg `solver_step!`, generic `Gradient` functor) — unsupported calls raise proper `MethodError`s, so `hasmethod`/`applicable` report the truth.
* Deleted dead code (`RealOrComplex`, `mat_x_vec!`, unused `pivots` cache field, NaN pre-fills, …); ~30 docstring corrections; line-search trial evaluations no longer write into the **shared solver cache** (aliasing hazard).
* Taxonomy: `LinesearchMethod <: SolverMethod` directly; `NonlinearSolverMethod` exported in place of the misleading `NonlinearMethod`.

## Interface-consistency audit (Phase 7)

A dedicated audit verified that all solvers within each category retain a consistent interface (source audit + a runtime `hasmethod`/construction/solve matrix over every concrete type). Line searches and nonlinear solvers were uniform; fixes applied where they weren't:

* `NonlinearSolver(QuasiNewton(n), …)` **silently discarded `n`** (always used the default 5) — the method's `refactorize` field is now honored.
* `solve(::LinearSolver, …)` added (only `solve(::LU, …)` existed — a prebuilt/pre-factorized solver could not be used non-mutatingly); the long-documented `solve!(x, lsolver, b)` got an actual LU implementation (it unconditionally threw "no method implemented").
* **`LinearProblem(A, y)` now stores copies of `A` and `y`** instead of NaN-initializing them — a freshly constructed problem is usable without `update!` (previously `solve(LU(), LinearProblem(A, b))` returned `[NaN, NaN]`). ⚠ Breaking; size-only constructors still allocate NaNs.
* `Quadratic`/`BierlaireQuadratic` constructors validate their parameters (like `Backtracking`/`StrongWolfe`); `DogLegSolver(x, y; F=missing)` uses the same friendly-error pattern as the other solvers.

## Verification

Beyond the per-fix regression tests, an independent verification pass re-exercised **every originally runtime-verified bug with fresh reproduction scripts** (30 checks) and audited every claimed fix against the source. It caught and fixed two issues in the remediation itself: the DogLeg trust radius was not reset by `initialize!` (state leak on solver reuse), and the NaN-recovery rescaling had become a latent infinite-loop hazard in the new ρ update.

**Quality gates** now in the suite: Aqua.jl (clean, including ambiguities), JET.jl (diagnostic), and a construct-every-export smoke test — the classes of tooling that would have caught most of §1 of the review up front.

**Test suite: 181 → 465 tests**, all passing (Smoke 59, Aqua 11, JET 1, Gradients 24, Jacobians 31, Nonlinear Problems 11, Hessians 8, Linear Solvers 68, Line Searches 127, Nonlinear Solvers 119, Newton/DogLeg 6). Doctests green.

## Breaking changes (⇒ 0.9.0)

See the **Removed/Changed (breaking)** sections of `CHANGELOG.md`: the removed dead-on-arrival exports (plus `minimum`, dropped with the optimizer subsystem that moved to a separate package), the **renamed nonlinear-solver method types** (`NewtonMethod` → `Newton`, `QuasiNewtonMethod` → `QuasiNewton` with `QuasiNewton = Newton{false}`, `PicardMethod` → `Picard`), removed phantom type parameters and size-only `NonlinearProblem` constructors, `LUSolverCache` field change, `NonlinearMethod` → `NonlinearSolverMethod`, the `LU()` `MMatrix` default, DogLeg's removed `refactorize` kwarg and new carried, `Options`-configurable ρ-based trust radius (`dogleg_radius_{initial,shrink,expand,max}`), Picard's fixed-point rework and rejected `linesearch` kwarg, `alloc_*` integer rejection, `LinearProblem` copy semantics, and quadratic constructor validation.

## Updates since the review

* The `check_jacobian`/`print_jacobian` solver forwarders (flagged as broken above) are now **fixed** and, together with `check_gradient`/`check_hessian`, take an optional leading `io::IO` argument; `print_jacobian` renders via `show(io, "text/plain", J)` (deterministic, io-honouring) instead of `display(J)`.
* The concrete nonlinear-solver method types were **renamed** to drop the `Method` suffix (`Newton`/`QuasiNewton`/`Picard`), so the method taxonomy is now naming-consistent.
* The DogLeg trust-region parameters are now `Options` fields (`dogleg_radius_{initial,shrink,expand,max}`), and DogLeg recovers from a collapsed (underflowed) trust radius instead of freezing.

## Known follow-ups (out of scope)

* A "pluggable linear solver" cleanup: generic `factorize!` fallbacks, error fallbacks on `AbstractLinearSolver`, exporting the `LinearSolverMethod`/`LinearSolverCache` extension points.
* Naming: `directions!` vs `direction!` — breaking rename deferred.
* Doctests run via the docs build, not `runtests.jl`.


test/aqua_tests.jl
```diff
@@ -0,0 +1,22 @@
...
+# Two checks are known to fail against the current (pre-Phase-1) source and are
+# marked `broken = true` (the Aqua analogue of `@test_broken`), so the overall
+# suite stays green while still enforcing the gate: once the corresponding fix
+# lands the check turns into an "Unexpected Pass", forcing the flip to
+# `broken = false`.
+#   * undefined exports  (§1.1 — `BracketingMethod`, `IterativeMethod`, and other
+#                          exported-but-undefined names) — fixed in Phase 1.1.
+#   * `convert` ambiguity (§1.5 — `Base.convert(::Type, ::LinesearchMethod)`)     —
+#                          fixed in Phase 1.4.
+# All other checks (stale deps, compat bounds, piracy, …) must pass now.
```
benedict-96: I'm not sure if the comments above are a good idea. They seem to be hard to read and difficult to maintain and overall do not add much value.

test/jet_tests.jl
```diff
@@ -0,0 +1,19 @@
+# JET.jl static analysis (Phase 0.4, optional / allowed-to-warn).
+#
+# `report_package` surfaces potential runtime errors detected statically.  The
+# current source has known issues (see `bugs.md`), so this set only reports and
+# is not allowed to fail the suite — it is a diagnostic aid for Phases 1–4.
+
```
benedict-96, line 4: Same as for the comments before `using Aqua`. Perhaps a link to the repo would be more instructive?

src/base/jacobian.jl
```diff
@@ -134,7 +123,7 @@ struct JacobianAutodiff{T,FT<:Callable,JT<:ForwardDiff.JacobianConfig,YT<:Abstra
     ty::YT
 
     function JacobianAutodiff(F::CT, x::YT, y::YT) where {T,YT<:AbstractArray{T},CT<:Callable}
-        applicable(F, y, x, ()) || error("The function needs to have the following signature: F(y, x, params).")
+        hasmethod(F, Tuple{typeof(y),typeof(x),Any}) || error("The function needs to have the following signature: F(y, x, params).")
```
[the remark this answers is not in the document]
michakraus: Fixed in 617fe45 — confirmed at runtime: `hasmethod(F, Tuple{typeof(y), typeof(x), Any})` only matches methods accepting *arbitrary* params, so `F(y, x, params::MyParams)` was rejected. The check now uses `methods(F, Tuple{typeof(y), typeof(x), Any})` (type-intersection matching), which accepts any 3-argument method matching `(y, x)` while still rejecting functions without a params argument. Regression test added.

src/nonlinear/nonlinear_solver.jl
```diff
@@ -29,7 +29,7 @@ struct NonlinearSolver{T,MT<:NonlinearSolverMethod,AT,NLST<:NonlinearProblem{T},
     config::Options{T}
 
     function NonlinearSolver(x::AT, nlp::NLST, ls::LST, linearsolver::LSoT, linesearch::LiSeT, cache::CT, config::Options{T}; method::MT=NewtonMethod(), jacobian::JT=JacobianAutodiff(nlp.F, x), options_kwargs...) where {T,AT<:AbstractVector{T},MT<:NonlinearSolverMethod,JT<:Jacobian,NLST,LST,LSoT,LiSeT,CT}
-        new{T,MT,AT,NLST,LST,JT,LSoT,LiSeT,CT}(nlp, ls, jacobian, linearsolver, linesearch, method, cache, config)
+        new{T,MT,NLST,LST,JT,LSoT,LiSeT,CT}(nlp, ls, jacobian, linearsolver, linesearch, method, cache, config)
```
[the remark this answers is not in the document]
michakraus: Fixed in 617fe45 — the inner constructor's `where` clause now mirrors the struct's parameter constraints (`JT<:Jacobian{T}`, plus the previously unconstrained `NLST`/`LST`/`LSoT`/`LiSeT`/`CT`), so an element-type mismatch fails with a clean `MethodError` at the constructor instead of a `TypeError` from `new{…}`.

src/base/gradient.jl
```diff
@@ -1,13 +1,27 @@
...
+1.1920928955078125e-7
+```
+
+```jldoctest; setup = :(using SimpleSolvers: default_ϵ)
+julia> default_ϵ(Float32)
+0.0027621358f0
+```
 """
-const DEFAULT_GRADIENT_ϵ = 8sqrt(eps())
+default_ϵ(::Type{T}) where {T<:Number} = 8sqrt(eps(T))
```
benedict-96, line 24: Does it make sense to rename `DEFAULT_GRADIENT_ϵ` to `default_ϵ`. I prefer the former as this provides more context ...

src/base/gradient.jl
```diff
@@ -98,7 +108,7 @@ function GradientFunction(F::Callable, ∇F!::Callable, x::AbstractVector{T}) wh
     GradientFunction{T}(F, ∇F!, length(x))
 end
 
-(grad::GradientFunction{T})(g::VT, x::VT) where {T,VT<:AbstractVector{T}} = grad.∇F!(g, x)
+(grad::GradientFunction{T})(g::AbstractVector{T}, x::AbstractVector{T}) where {T} = grad.∇F!(g, x)
```
benedict-96, line 115: For the Euclidean gradient it makes sense that `g` and `x` are of the same type. I would not include this change.

src/base/jacobian.jl
```diff
@@ -1,14 +1,3 @@
-"""
-    DEFAULT_JACOBIAN_ϵ
-
-A constant used for computing the finite difference Jacobian. See [`JacobianFiniteDifferences`](@ref).
-
-# Extended help
-
-For the [`GradientFiniteDifferences`](@ref) this is called [`DEFAULT_GRADIENT_ϵ`](@ref).
-"""
-const DEFAULT_JACOBIAN_ϵ = 8sqrt(eps())
```
benedict-96, line 10: This concerns the earlier comment made for `default_ϵ` (in the gradient case). Should we have separate default $\epsilon$ for the gradient and the Jacobian?

src/base/jacobian.jl
```diff
@@ -125,7 +114,7 @@ JacobianAutodiff{T}(F, nx::Integer)
 The functor does:
 
 ```julia
-jac(J, x) = ForwardDiff.jacobian!(J, jac.ty, x, grad.Jconfig)
+jac(J, x, params) = ForwardDiff.jacobian!(J, (y, x) -> jac.F(y, x, params), jac.ty, x, jac.Jconfig)
```
benedict-96, line 138: Good catch!
But should we turn the code chunk into an executable `jldoctest`?

src/base/jacobian.jl
```diff
@@ -134,7 +123,13 @@ struct JacobianAutodiff{T,FT<:Callable,JT<:ForwardDiff.JacobianConfig,YT<:Abstra
     ty::YT
 
     function JacobianAutodiff(F::CT, x::YT, y::YT) where {T,YT<:AbstractArray{T},CT<:Callable}
-        applicable(F, y, x, ()) || error("The function needs to have the following signature: F(y, x, params).")
+        # Accept any three-argument method whose first two arguments match (y, x).
+        # `hasmethod(F, Tuple{…,Any})` would be too strict: it only matches methods
+        # accepting *arbitrary* params, spuriously rejecting a valid
+        # `F(y, x, params::MyParams)`.  `methods` matches by type intersection, so
+        # typed-params methods are found while two-argument functions are still
+        # rejected.
```
benedict-96: I think this series of comments is not really that helpful. I just adds overhead that needs to be maintained. The same statements could be included in the form of tests?

src/bracketing/bracket_minimum.jl
```diff
@@ -67,13 +67,16 @@ struct BracketRootCriterion <: BracketingCriterion end
...
     b = a + s
     yb = f(b)
 
-    # check if condition is already satisfied
-    if bc(f(a - s), yb)
+    # Check if the condition is already satisfied to the left of `a`.  This early
+    # exit is only valid for the *root* criterion (a sign change between `a - s`
+    # and `b`).  For the *minimum* criterion it would bracket a maximum: under the
+    # invariant f(b) ≤ f(a) the interior point `b` is the highest of the three, so
+    # `f(a - s) ≥ f(b)` does not indicate a minimum in `(a - s, b)`.
```
benedict-96: Again, a lot of information in the form of hard-to-maintain in-line comments. This may be better placed in an `# Extended hlep` section of the associated docstring?

src/bracketing/bracket_minimum.jl
```diff
@@ -248,5 +250,5 @@ function bracket_root(f::Callable, x::T; s::T=T(DEFAULT_BRACKETING_s), k::T=T(DE
 end
 
 function bracket_root(prob::LinesearchProblem{T}, params, x::T; kwargs...) where {T<:Number}
-    bracket_root(x -> value(prob, x, params), x -> derivative(prob, x, params), x; kwargs...)
+    bracket_root(β -> value(prob, β, params), x; kwargs...)
```
benedict-96, line 290: The derivative is not needed anymore?

src/bracketing/triple_point_finder.jl
```diff
@@ -30,26 +30,38 @@ julia> round10.((f(a), f(b), f(c)))
 The algorithm is taken from [bierlaire2015optimization; Chapter 11.2.1](@cite).
 """
 function triple_point_finder(f::Callable, x₀::T, δ::T, nmax::Integer=DEFAULT_BRACKETING_nmax, adjust_constant_iteration::Integer=1) where {T}
+    # Evaluate `f` once per point and carry the values across iterations: each
+    # `f(xₖ)` was previously the `f(xₖ₊₁)` of the prior iteration, so caching
+    # roughly halves the number of function evaluations.
```
benedict-96: Again, do we need inline comments?

src/bracketing/triple_point_finder.jl
```diff
@@ -30,26 +30,38 @@ julia> round10.((f(a), f(b), f(c)))
...
+    fx₀ = f(x₀)
     x₁ = x₀ + δ
+    fx₁ = f(x₁)
 
-    if f(x₁) ≥ f(x₀)
+    if fx₁ ≥ fx₀
```
benedict-96, line 44: Does this change make sense? `f` anyway is just a scalar function ...

src/bracketing/triple_point_finder.jl
```diff
@@ -30,26 +30,38 @@ julia> round10.((f(a), f(b), f(c)))
...
         if adjust_constant_iteration > MAX_NUMBER_ADJUST_CONSTANT_ITERATIONS
-            error("The function `f` must be decreasing at `$(x₀)`; `f($(x₁)) = $(f(x₁))` must be smaller than `f($(x₀)) = $(f(x₀))`.")
+            error("The function `f` must be decreasing at `$(x₀)`; `f($(x₁)) = $(fx₁)` must be smaller than `f($(x₀)) = $(fx₀)`.")
         end
-        triple_point_finder(f, x₀, δ / 2, nmax, adjust_constant_iteration + 1)
+        return triple_point_finder(f, x₀, δ / 2, nmax, adjust_constant_iteration + 1)
```
benedict-96, line 48: I guess that's a question of style. I prefer not having the `return` at the end of the function ...

src/bracketing/triple_point_finder.jl
```diff
@@ -30,26 +30,38 @@ julia> round10.((f(a), f(b), f(c)))
...
     end
 
+    # Initialize so that the first iteration sees xₖ₋₁ = x₀, xₖ = x₁ (rather than
+    # the degenerate xₖ₋₁ = xₖ = x₁, which would return two identical points and
+    # make the downstream quadratic fit singular).
```
benedict-96: Could catch/fix, but unnecessary comment (in my opinion).

src/linear/linear_problem.jl
```diff
@@ -29,7 +29,9 @@ LinearProblem{T}(n, m)
 LinearProblem{T}(n)
 ```
 
-Note that in any case the allocated system is initialized with `NaN`s:
+`LinearProblem(A, y)` stores *copies* of `A` and `y`, so the problem is ready to
+solve right after construction (and later changes to the caller's arrays do not
+leak into the problem):
```
benedict-96: Somewhat weird phrase. Does that mean that it's simply immutable?

src/linear/linear_solvers.jl
```diff
@@ -43,7 +43,7 @@ Return the cache of the [`LinearSolver`](@ref).
 julia> ls = LinearSolver(LU(), [1 2; 3 4]);
 
 julia> cache(ls)
-SimpleSolvers.LUSolverCache{Int64, StaticArraysCore.MMatrix{2, 2, Int64, 4}}([1 2; 3 4], [0, 0], [0, 0], 0)
+SimpleSolvers.LUSolverCache{Float64, Matrix{Float64}}([1.0 2.0; 3.0 4.0], [0, 0], 0)
```
benedict-96: This was an `MMatrix` before and is a `Matrix` now. Does that make sense?

src/linear/lu_solver.jl
```diff
@@ -57,50 +56,55 @@ struct LU{ST<:Union{Missing,Bool}} <: DirectMethod
     LU(; pivot=true, static=missing) = new{typeof(static)}(static, pivot)
 end
 
-"""
-Threshold for the maximum size a static matrix should have. See [`_static`](@ref).
-"""
-const N_STATIC_THRESHOLD = 10
```
benedict-96, line 63: The `static` threshold has been completely removed?

src/linear/lu_solver.jl
```diff
@@ -57,50 +56,55 @@ struct LU{ST<:Union{Missing,Bool}} <: DirectMethod
...
 
 # Keys
 - `A`: the factorized matrix `A`,
-- `pivots`: a vector of pivots used during factorization,
 - `perms`: a vector of permutations used during factorization,
 - `info`: stores an index regarding pivoting.
 """
 mutable struct LUSolverCache{T,AT<:AbstractMatrix{T}} <: LinearSolverCache{T}
     A::AT
-    pivots::Vector{Int}
```
benedict-96, line 89: We don't need `pivots` anymore?

src/linear/lu_solver.jl
```diff
@@ -57,50 +56,55 @@ struct LU{ST<:Union{Missing,Bool}} <: DirectMethod
...
+type that supports division (mirroring `LinearAlgebra.lutype`), so that e.g.
+`LinearSolver(LU(), [1 2; 3 4])` works instead of failing in `factorize!`.
+"""
+lucache_eltype(::Type{T}) where {T} = typeof(oneunit(T) / oneunit(T))
+
+# Build the cache matrix, choosing static vs. dynamic storage.  The size of a
+# `StaticMatrix` is part of its *type*, so `MMatrix{M,N,Tf}(A)` is fully
+# inferrable; for a general `AbstractMatrix` the size is only known at runtime,
+# so we allocate a plain `Matrix`.  Keeping these on separate methods (rather
+# than a runtime `?:`) makes the default `LinearSolverCache` path type stable.
```
benedict-96: Again, the inline comments are not too instructive and difficult to maintain. Perhaps include that information in the respective docstring?

src/linear/lu_solver.jl
```diff
@@ -57,50 +56,55 @@ struct LU{ST<:Union{Missing,Bool}} <: DirectMethod
...
 end
 
 function LinearSolverCache(lu::LU{Bool}, A::AbstractMatrix{T}) where {T}
     n = checksquare(A)
-    Ā = lu.static ? MMatrix{size(A)...}(copy(A)) : copy(A)
-    LUSolverCache{T,typeof(Ā)}(Ā, zeros(Int, n), zeros(Int, n), 0)
+    Tf = lucache_eltype(T)
+    # An explicit `static` keyword overrides the dispatch default.  `static=true`
+    # on a dynamically-sized matrix is inherently a runtime-sized `MMatrix` (opt-in,
+    # built once at construction); the default `LU()` path above stays type stable.
```
benedict-96: Again, information is probably (if included at all) better suited to the docstring. The comment with the `static` keyword also seems confusing? There is a `static` field in `lu` but no keyword specified ...

src/linear/lu_solver.jl
```diff
@@ -111,9 +115,17 @@ function solve!(solution::AbstractVector, lsolver::LinearSolver{T,LUT}, ls::Line
 end
 
 function solve!(solution::AbstractVector, lsolver::LinearSolver{T,LUT}, A::AbstractMatrix, b::AbstractVector) where {T,LUT<:LU}
-    ls = LinearProblem(solution)
-    update!(ls, A, b)
-    solve!(solution, lsolver, ls)
+    # Copy `A` straight into the existing cache and factorize in place, rather than
+    # allocating a throwaway `LinearProblem` on every call.
```
benedict-96: We should get of this comment I think ...

src/linear/lu_solver.jl
```diff
@@ -150,17 +162,16 @@ julia> b = ones(3)
  1.0
 
 julia> solve(LU(), A, b)
-3-element StaticArraysCore.SizedVector{3, Float64, Vector{Float64}} with indices SOneTo(3):
+3-element Vector{Float64}:
```
benedict-96: Why is this not a static array anymore. Has the default changed?
That probably has to do with the fact that the threshold constant (for $n \leq 10$) was removed ...

src/linear/lu_solver.jl
```diff
@@ -150,17 +162,16 @@ julia> b = ones(3)
...
  0.5
  0.25
 ```
 Compare this to [`solve!(::AbstractVector, ::LinearSolver, ::LinearProblem)`](@ref).
 """
 function solve(lu::LU, A::AbstractMatrix, b::AbstractVector)
-    ls = LinearProblem(A, b)
-    update!(ls, A, b)
-    solve(lu, ls)
+    # `LinearProblem(A, b)` stores copies of `A` and `b`, so no `update!` is needed.
```
benedict-96: Good catch! But this comment shouldn't be there 🥴

src/linear/lu_solver.jl
```diff
@@ -203,27 +214,34 @@ Here `cache(lsolver).A` stores the factorized matrix. If we call `factorize!` wi
 julia> lsolver = LinearSolver(LU(), A);
 
 julia> factorize!(lsolver).cache.A
-3×3 StaticArraysCore.MMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3):
+3×3 Matrix{Float64}:
```
benedict-96: Again, this should be a static array, right?

src/linear/lu_solver.jl
```diff
@@ -203,27 +214,34 @@ Here `cache(lsolver).A` stores the factorized matrix. If we call `factorize!` wi
...
 ```
 
-Also note the difference between the output types of the two refactorized matrices. This is because we set the keyword `static` to false when calling [`LU`](@ref). Also see [`_static`](@ref).
+For a plain (dynamically-sized) matrix, both the default `LU()` and `LU(; static=false)` allocate a plain `Matrix` cache. Pass a `StaticArray` (or `LU(; static=true)`) to obtain a mutable static (`MMatrix`) cache instead.
 
 Also see [`ldiv!`](@ref) for how the refactorized matrix is used.
 """
 function factorize!(lsolver::LinearSolver{T,LUT}) where {T,LUT<:LU}
+    # The hand-rolled factorization below indexes `1:n` under `@inbounds`, so it is
+    # only correct for one-based storage.
```
benedict-96: This comment should probably not be here.

src/linear/lu_solver.jl
```diff
@@ -203,27 +214,34 @@ Here `cache(lsolver).A` stores the factorized matrix. If we call `factorize!` wi
...
+    Base.require_one_based_indexing(cache(lsolver).A)
+
+    # Reset the singularity marker before (re)factorizing so that a stale nonzero
+    # `info` from a previous factorization does not persist.
```
benedict-96: This comment should go as well.

src/linear/lu_solver.jl
```diff
@@ -263,13 +281,15 @@ end
 factorize!(lsolver::LinearSolver{T,LUT}, ls::LinearProblem{T}) where {T,LUT<:LU} = factorize!(lsolver, ls.A)
 
 """
-    find_maximum_value(v, k)
+    pivot_index(v, k)
```
benedict-96, line 292: I'm not sure if `pivot_index` is better than `find_maximum_value`? The former is more descriptive I think ...

src/linear/lu_solver.jl
```diff
@@ -263,13 +281,15 @@ end
...
 
-Find the maximum value of vector `v` starting from the index `k`.
+Return the index (starting from `k`) of the entry of `v` with the largest absolute
+value.
 
 This is used for *pivoting* in [`factorize!`](@ref).
 """
-function find_maximum_value(v::AbstractVector{T}, k::Integer) where {T<:Number}
+function pivot_index(v::AbstractVector{T}, k::Integer) where {T<:Number}
```
benedict-96, line 299: Same comment as above.

src/linear/lu_solver.jl
```diff
@@ -322,8 +342,22 @@ julia> ldiv!(x, s, b)
 function LinearAlgebra.ldiv!(x::AbstractVector{T}, lsolver::LinearSolver{T,LUT}, b::AbstractVector{T}) where {T,LUT<:LU}
     @assert axes(x, 1) == axes(b, 1) == axes(cache(lsolver).A, 1) == axes(cache(lsolver).A, 2)
 
+    # The substitution loops below index `1:n` under `@inbounds`, so they are only
+    # correct for one-based storage.
```
benedict-96: This comment can be left out.

src/linear/lu_solver.jl
```diff
@@ -322,8 +342,22 @@ julia> ldiv!(x, s, b)
...
+    Base.require_one_based_indexing(x, b, cache(lsolver).A)
+
+    # A zero pivot was encountered during factorization: the matrix is singular
+    # and back-/forward-substitution below would silently produce NaN/Inf.
```
benedict-96: Comments don't make much sense here.

src/linesearch/curvature_condition.jl
```diff
@@ -19,27 +19,35 @@ With the strong curvature condition we check:
...
 end
 
 function (cc::CurvatureCondition{T,DT,:Standard})(α::T) where {T,DT}
-    cc.D(α) ≥ cc.c * cc.d
+    cc.D(α) ≥ cc.c * cc.d₀
 end
 
 function (cc::CurvatureCondition{T,DT,:Strong})(α::T) where {T,DT}
-    abs(cc.D(α)) < abs(cc.c * cc.d)
+    abs(cc.D(α)) ≤ abs(cc.c * cc.d₀)
```
benedict-96, line 52: I think in Nocedal/Wright they have the strong inequality.

src/linesearch/backtracking.jl
```diff
@@ -20,10 +20,11 @@ const DEFAULT_ARMIJO_p = 0.5
 @doc raw"""
     const DEFAULT_WOLFE_c₁
 
-A constant ``c_1`` that is used in the [`SufficientDecreaseCondition`](@ref):
+A constant ``c_1`` that is used in the [`SufficientDecreaseCondition`](@ref) (the
+Armijo condition):
 
 ```math
-\frac{f(\alpha) - f(\alpha_0)}{c_1} < \alpha\cdot{}f'(\alpha_0).
+f(\alpha) \leq f(\alpha_0) + c_1 \alpha f'(\alpha_0).
```
benedict-96, line 27: Again, I think that standard textbooks have a strong inequality here. This change is not really necessary.

src/linesearch/backtracking.jl
```diff
@@ -106,22 +107,45 @@ function solve(ls::Linesearch{T,<:Backtracking}, α::T, params=NullParameters())
...
     sdc = SufficientDecreaseCondition(method(ls).c₁, y₀, d₀, f)
-    cc = CurvatureCondition(method(ls).c₂, d₀, d; mode=:Standard)
 
+    # Textbook backtracking (Nocedal & Wright, Alg. 3.1) terminates on the
+    # sufficient-decrease (Armijo) condition alone.  Enforcing the curvature
+    # condition inside the shrink-only loop below can never terminate, because
+    # shrinking α makes the curvature condition harder to satisfy (α → 0 ⇒
+    # D(α) → d₀ < c₂·d₀).  The curvature condition is therefore checked only
+    # post-hoc (see below), not as a loop termination criterion.
+    αₐ = α₀  # last α satisfying sufficient decrease (α₀ = 0 trivially does)
```
benedict-96: Good catch! But we should remove the comment. This won't make any sense after the pr is merged ...

src/linesearch/backtracking.jl
```diff
@@ -106,22 +107,45 @@ function solve(ls::Linesearch{T,<:Backtracking}, α::T, params=NullParameters())
...
+        # Never silently return a denormal α: fall back to the last α that
+        # satisfied sufficient decrease (α₀ = 0 if none did).
+        config(ls).verbosity ≥ 1 && @warn "Backtracking line search did not satisfy the sufficient decrease condition within $(config(ls).max_iterations) iterations. Returning α = $(αₐ)."
+        return αₐ
+    end
+
+    # Opt-in, post-hoc curvature (second Wolfe) check.  If the accepted step
+    # fails it we warn rather than shrinking further (which would break the
+    # sufficient decrease guarantee); enforcing curvature requires a proper
+    # bracketing/zoom line search.
```
benedict-96: Again the comment is unnecessary and should be removed. But in light of that, should we just get rid of the curvature conditions when using backtracking? We may also have to investigate this ...
I remember Nocedal/Wright using the curvature condition for backtracking as well.

src/linesearch/bisection.jl
```diff
@@ -38,10 +38,14 @@ So the algorithm checks in each step where the sign change occurred and moves th
...
     The obvious danger with using bisections is that the supplied interval can have multiple roots (or no roots). One should be careful to avoid this when fixing the interval.
 """
-function bisection(f::Callable, αmin::T, αmax::T, params=NullParameters(), config::Options=Options(T)) where {T<:Number}
-    α₀ = αmin
-    α₁ = αmax
-    α = zero(T)
+function bisection(f::Callable, αmin::T, αmax::T, params=NullParameters(), config::Options=Options(float(T))) where {T<:Number}
+    # Promote to a floating point type on entry: with an integer `T` the midpoint
+    # `(α₀ + α₁) / 2` would silently switch `α`'s type mid-loop, and `Options(T)`
+    # is undefined for integer `T`.
```
benedict-96: Again, this comment will most likely become irritating garbage at some point in the future. Should best be removed.

src/linesearch/bisection.jl
```diff
@@ -52,14 +56,28 @@ function bisection(f::Callable, αmin::T, αmax::T, params=NullParameters(), con
...
     y = zero(y₀)
 
-    # @assert y₀ * y₁ ≤ 0 "Either no or multiple real roots in [xmin,xmax]."
+    # Bisection can only locate a root if the endpoints straddle a sign change.
+    # Same-sign endpoints mean there is no (odd-multiplicity) root in the interval;
+    # the loop below would otherwise silently collapse onto α₁ and return a
+    # non-root.  This case arises benignly in the line search once the derivative
+    # has flattened at a minimum (both endpoint values ≈ 0 with the same sign), so
+    # rather than erroring we return the endpoint closest to a root (smallest |f|)
+    # and warn only at high verbosity.
```
benedict-96: Good and important/essential point! But this comment is better included as part of the docstring than having it here in the source code.

src/linesearch/bisection.jl
```diff
@@ -69,16 +87,29 @@ function bisection(f::Callable, αmin::T, αmax::T, params=NullParameters(), con
...
 
-bisection(f::Callable, α::T, params=NullParameters(), config::Options=Options(T)) where {T<:Number} = bisection(f, bracket_minimum(f, α)..., params, config)
+bisection(f::Callable, α::T, params=NullParameters(), config::Options=Options(float(T))) where {T<:Number} = bisection(f, bracket_root(β -> f(β, params), α)..., params, config)
+
+# Disambiguation: `(f, ::T, ::T, ::Options)` matches both the interval form above
+# (with the `Options` landing in the untyped `params` slot) and the single-`α`
+# convenience form (with the second `T` landing in its untyped `params` slot),
+# and neither is more specific than the other.  A call with two numeric bounds and
+# an `Options` is unambiguously the interval form with default `params`, so route
+# it there and give the second `T` bound priority over the single-`α` reading.
```
benedict-96: I guess this also falls into the category of "rambling in inline comments in the source code"

src/linesearch/bisection.jl
```diff
@@ -96,14 +127,18 @@ function solve(ls::Linesearch{T,<:Bisection}, α₀::T, α₁::T, params=NullPar
 end
 
 function solve(ls::Linesearch{T,<:Bisection}, α::T, params=NullParameters()) where {T}
-    # TODO: The following line should use α instead of zero(T) but that requires a rework of the bracketing algorithm
```
benedict-96, line 99: This TODO seems to not have been resolved!

src/linesearch/bisection.jl
```diff
@@ -96,14 +127,18 @@ function solve(ls::Linesearch{T,<:Bisection}, α₀::T, α₁::T, params=NullPar
...
-    # solve(problem, ls, bracket_minimum(problem.F, α)..., params)
+    # Design note (Phase 5, resolving the former "use α" TODO): the bracket is
+    # anchored at α = 0 rather than at the caller's α.  Starting `bracket_minimum`
+    # at α is unsafe — the merit need not be decreasing there and the search could
+    # wander to a negative (infeasible) step — and using α as the bracketing step
+    # size over-coarsens the search and destabilises stiff problems.  The step
+    # magnitude is governed by `bracket_minimum`'s tuned default, not by α.
```
benedict-96: Ok, so this we should think about. That's a valid point but having an iterative way of incorporating new $\alpha$s would probably be good.

src/linesearch/linesearch.jl
```diff
@@ -1,21 +1,35 @@
...
+    LinesearchMethod{T} <: SolverMethod
 
 Examples include [`Static`](@ref), [`Backtracking`](@ref), [`Bisection`](@ref) , [`BierlaireQuadratic`](@ref) and [`Quadratic`](@ref).
 See these examples for specific information on linesearch algorithms.
 
 # Extended help
 
 A `LinesearchMethod` is usually used in [`Linesearch`](@ref) (or with [`solve`](@ref)).
+
+It is a subtype of `SolverMethod` (imported from `GeometricBase`) — line searches
```
benedict-96, line 11: Just a side note: we could link docstrings from other packages here.

src/linesearch/quadratic.jl
```diff
@@ -57,49 +58,66 @@ function Quadratic(::Type{T}=Float64;
...
-    # solve(ls, α₀, params, method(ls).s, 0)
-    solve(ls, zero(T), params, method(ls).s, 0)
+    # Design note (Phase 5, resolving the former "use α₀" TODO): the caller's α₀ is
+    # deliberately *not* used as the bracket start.  `bracket_minimum_with_fixed_point`
+    # holds its left endpoint fixed and only expands to the right, so it must start
+    # where the merit is guaranteed to be decreasing — α = 0 (φ'(0) < 0 for a
+    # descent direction).  Anchoring at α₀ > 0 would fail whenever the optimal step
+    # is smaller than α₀, and using α₀ as the initial step *size* over-coarsens the
+    # bracket and destabilises stiff problems (the tuned `method(ls).s` is required).
+    # The step magnitude is therefore governed by the method's `s`, not by α₀.
```
benedict-96: Ok, good point. This also related to the comments above https://github.com/JuliaGNI/SimpleSolvers.jl/pull/161/changes#r3557944024 and https://github.com/JuliaGNI/SimpleSolvers.jl/pull/161/changes#r3557956502 but I'll have to look through this in some detail. Claude seems to have solved this.
The rambling in inline comments in the source code is highly irritating though.

src/linesearch/quadratic.jl
```diff
@@ -57,49 +58,66 @@ function Quadratic(::Type{T}=Float64;
...
+        # compute minimum αₜ of p(α); i.e. p'(α) = 0.
+        # αₜ = a - p₁ / (2p₂)
+
+        # The denominator is 2·p₂·(b - a)², proportional to the fitted curvature p₂.
+        # If p₂ ≤ 0 (locally linear or non-convex fit) the quadratic model has no
+        # interior minimum; if the resulting αₜ is not finite the fit is degenerate.
+        # In either case fall back to a bisection step of the current bracket [a, b]
+        # instead of producing an Inf/NaN αₜ.  (Note: p₂ is small but positive near
+        # convergence, where the interpolation is still valid, so we guard on the sign
+        # and finiteness rather than on a magnitude threshold.)
```
benedict-96: Some comments above were here before, some newly introduced. This has to be looked at in detail.

src/linesearch/quadratic_bierlaire.jl
```diff
@@ -19,23 +19,10 @@ julia> default_precision(Float32)
...
-
-function default_precision(::Type{Float32})
-    8eps(Float32)
-end
-
-function default_precision(::Type{Float64})
-    8eps(Float64)
-end
-
-function default_precision(::Type{T}) where {T<:AbstractFloat}
```
benedict-96, line 36: There was no default precision defined for `Float16` on purpose. But it may make sense to support this ...

src/linesearch/wolfe.jl
```diff
@@ -0,0 +1,160 @@
...
+- `c₂` (default [`DEFAULT_WOLFE_c₂`](@ref)): the curvature constant ``c_2``. We require ``c_1 < c_2 < 1``.
+- `αmax` (default [`DEFAULT_WOLFE_αmax`](@ref)): the largest step the bracketing phase will try.
+
+!!! info
+    The strong Wolfe conditions require a *descent direction* (``f'(0) < 0``).  If
+    the line search problem is not decreasing at ``\alpha = 0`` the method cannot
+    make progress; it then returns the caller's initial step and (at
+    `verbosity ≥ 1`) warns.
+"""
+struct StrongWolfe{T} <: LinesearchMethod{T}
```
benedict-96, line 46: Are the strong Wolfe conditions not just the sufficient decrease condition + the strong curvature condition? This should, if we define it at all, use these constructions to avoid duplicate code.

src/nonlinear/newton_solver.jl
```diff
@@ -129,5 +136,9 @@ QuasiNewtonSolver(args...; kwargs...) = NewtonSolver(args...; refactorize=DEFAUL
 check_jacobian(s::Union{NewtonSolver,QuasiNewtonSolver}) = check_jacobian(jacobian(s))
 print_jacobian(s::Union{NewtonSolver,QuasiNewtonSolver}) = print_jacobian(jacobian(s))
 
-NonlinearSolver(method::NewtonMethod, args...; kwargs...) = NewtonSolver(args...; kwargs...)
-NonlinearSolver(method::QuasiNewtonMethod, args...; kwargs...) = QuasiNewtonSolver(args...; kwargs...)
+# Honor the method's `refactorize` field (this covers `QuasiNewtonMethod` =
```
benedict-96: Very weird comment. Should be removed.

src/nonlinear/nonlinear_solver.jl
```diff
@@ -28,8 +28,12 @@ struct NonlinearSolver{T,MT<:NonlinearSolverMethod,AT,NLST<:NonlinearProblem{T},
     cache::CT
     config::Options{T}
 
-    function NonlinearSolver(x::AT, nlp::NLST, ls::LST, linearsolver::LSoT, linesearch::LiSeT, cache::CT, config::Options{T}; method::MT=NewtonMethod(), jacobian::JT=JacobianAutodiff(nlp.F, x), options_kwargs...) where {T,AT<:AbstractVector{T},MT<:NonlinearSolverMethod,JT<:Jacobian,NLST,LST,LSoT,LiSeT,CT}
-        new{T,MT,AT,NLST,LST,JT,LSoT,LiSeT,CT}(nlp, ls, jacobian, linearsolver, linesearch, method, cache, config)
+    # The `where` clause mirrors the struct's parameter constraints (notably
+    # `JT<:Jacobian{T}`), so an element-type mismatch (e.g. a `Jacobian{Float32}`
+    # with `T = Float64`) fails with a clean `MethodError` at the constructor
+    # instead of a confusing `TypeError` from `new{…}`.
```
benedict-96: Commenting the changes in this particular pr should not be in the source code.

src/nonlinear/nonlinear_solver.jl
```diff
@@ -84,7 +86,7 @@ linearsolver(s)
 
 # output
 
-LinearSolver{Float64, LU{Missing}, SimpleSolvers.LUSolverCache{Float64, StaticArraysCore.MMatrix{3, 3, Float64, 9}}}(LU{Missing}(missing, true), SimpleSolvers.LUSolverCache{Float64, StaticArraysCore.MMatrix{3, 3, Float64, 9}}([0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0], [0, 0, 0], [0, 0, 0], 0))
+LinearSolver{Float64, LU{Missing}, SimpleSolvers.LUSolverCache{Float64, Matrix{Float64}}}(LU{Missing}(missing, true), SimpleSolvers.LUSolverCache{Float64, Matrix{Float64}}([0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0], [0, 0, 0], 0))
```
benedict-96: Relates to a comment above. I think this should still be a static array.

src/nonlinear/nonlinear_solver_state.jl
```diff
@@ -53,10 +53,10 @@ function increase_iteration_number!(state::NonlinearSolverState)
 end
 
 function NonlinearSolverState{T}(n::Integer, m::Integer=n) where {T}
+    # The two-argument `NonlinearSolverState` already NaN-fills fresh copies of
+    # its inputs, so pre-filling `x`/`y` here would be redundant.
```
benedict-96: This comment is redundant.

src/utils.jl
```diff
@@ -19,13 +19,21 @@ Allocate `NaN`s of the size of the Hessian of `f` (with respect to `x`).
 """
 alloc_h
 
-alloc_x(x::Number) = typeof(x)(NaN)
+# The `alloc_*` helpers initialize their storage with `NaN`, which only exists for
+# floating-point (and complex-of-floating-point) element types.  Passing an
+# integer array would otherwise fail deep inside the broadcast with a cryptic
+# `InexactError`; raise a clear error instead.
```
benedict-96: The integer array comment can be removed. This concerns current, pending, changes.

test/failing_newton_iterations.jl
```diff
@@ -13,13 +13,27 @@ root(::Type{T}) where {T} = zeros(T, 2)
 tol(::Type{T}) where {T} = T == Float64 ? eps(T) : eps(T)
 
 function try_different_solvers(T::DataType)
+    # NewtonSolver: with the Phase 2 fixes this now converges on the Powell
+    # problem.  Previously it *stagnated* at x ≈ [1.108, 0] and that stalled
+    # iterate was falsely reported as converged (bugs.md §3): the backtracking
+    # line search shrank α to a denormal (§1.3) and the successive-change
+    # convergence criteria treated the resulting zero step as convergence.
+    # Fixing the backtracking stall (2.1) and requiring a small residual for
+    # convergence (2.2) lets Newton escape the stagnation point and reach the
```
benedict-96: The changes that happened during this pr should not be commented in the form of inline comments in the source code (again).

test/failing_newton_iterations.jl
```diff
@@ -13,13 +13,27 @@ root(::Type{T}) where {T} = zeros(T, 2)
...
     solve!(x0, solver)
-    @test_throws AssertionError @assert ≈(x0, _root; atol=tol(T))
+    @test ≈(x0, _root; atol=tol(T))
 
+    # PicardSolver cannot solve this problem, but for a principled reason: since
+    # Phase 5 it is a proper (residual-safeguarded) fixed-point iteration
+    # x ← x + α(-F(x)), and the Powell map is not a contraction here, so it stalls
+    # at a non-root instead of converging.  Crucially it does *not* diverge to NaN
+    # or falsely report convergence (Phase 2 residual gate) — it simply runs out of
+    # iterations, so the equality assertion below fails as expected.
```
benedict-96: Again, these comments shouldn't be here.

benedict-96 (dismissed):
I checked the source code in detail and marked some changes that should be made. I also skimmed over the tests, but these look largely fine. Some of the more complicated algorithmic aspects I have yet to check in detail, but marked the relevant bits in my comments. For all those algorithmic changes it may make sense to split the pr in multiple smaller ones that would make it easier to handle.

src/base/gradient.jl
michakraus: As this is the same for gradient and jacobian, I think this is fine. There is not much value in having separate methods.

src/base/gradient.jl
michakraus: Potentially, we could have different wrappers here, that provide additional context, similar to `StateVariable` and `VectorField` in the Geometric* packages. Restricting the value and its gradient to the same array type is unnecessarily restrictive.

benedict-96: Alright, then I leave this as is 👍

src/base/jacobian.jl
michakraus: I think there is not really any added value in that.

src/base/jacobian.jl
michakraus: Probably makes sense.

src/bracketing/triple_point_finder.jl
michakraus: I guess this complies with the style guidelines.

src/linear/linear_problem.jl
michakraus: No, this means that if you change `A` and `y` after initialisation, this will not change the corresponding arrays in the `LinearProblem`.

src/linear/linear_solvers.jl
michakraus: This is one of the things I am not so sure about. I guess the test used static arrays intentionally.

src/linear/lu_solver.jl
michakraus: Yes, this is documenting changes, not documenting the code.

src/bracketing/bracket_minimum.jl
michakraus: I guess we have
```
    bracket_root(f, x)

Make a bracket for the function based on `x` (for root finding).
```
where `x` does not seem to be a derivative.

src/bracketing/triple_point_finder.jl
michakraus: I think this is to reuse `fx₁`and `fx₀` below without recomputing them.

src/linear/lu_solver.jl
michakraus: As the permutations are stored, we probably don't need to store the pivots as well. This is never used.

src/linear/linear_solvers.jl
michakraus: Ideally, we would have two doctests, one for static arrays and one for standard arrays, using the correct flags.

src/linear/lu_solver.jl
michakraus: I am not sure why this happened. I guess one can argue for no implicit type changes. If I request static, I get static. If I put in static, I get static out. But if I do not put in static and the matrix is small, I should not get static out.

src/linear/lu_solver.jl
michakraus: Cf. comment above.

PR #171, opened 2026-07-30
Standardise the line-search contract and make it robust at the merit's round-off floor

michakraus:
Fixes the `Backtracking` line-search warning that floods downstream packages
(GeometricIntegrators, GeometricProblems, ChargedParticleDynamics):

```
┌ Warning: Backtracking line search did not satisfy the sufficient decrease condition within 1000 iterations. Returning the last trial step α = 2.220446049250313e-16.
└ @ SimpleSolvers src/linesearch/backtracking.jl:127
```

Five commits: the fix itself, then a budget unification, then a standardised contract across
all six line searches (which fixed two `BierlaireQuadratic` defects found on the way), then two
rounds of fixes from reviewing those.

## Root cause

`2.220446049250313e-16 == eps(1.0) == 1.0·0.5^52`, so the ladder always exited via its
`α ≤ eps` guard after ~53 trials — **the iteration count in the message was never true**, and
neither was the implied diagnosis.

The sufficient decrease condition `φ(α) ≤ φ(0) + c₁·α·φ'(0)` demands a decrease *proportional
to* `α` with no round-off allowance. Once `c₁·α·|φ'(0)|` falls below one ulp of `φ(0)` — at
`α* = eps/(4c₁) ≈ 5.55e-13` for the `‖F‖²` merit of a Newton step, independent of problem
scale — the right-hand side rounds back up to `φ(0)` exactly and the accept/reject decision is
made by **rounding alone**. Two failure modes follow, and **the reported warning is the less
harmful one**:

| regime | behaviour before |
|---|---|
| trial point stops moving above `α*` | the condition **"succeeded"** at `α ≈ 2.3e-13`, purely because a frozen merit tied a rounded right-hand side. The returned step did not move `x` at all — **and nothing was reported** |
| `‖δx‖ ≳ 400‖x‖` (near-singular J), or `\|φ'(0)\| ≳ 2500·φ(0)` (inexact solve, stale quasi-Newton Jacobian, regularization) | the ladder ran to `α = eps` and emitted the warning above |

Either way the outer iterate froze, the solve spun to `max_iterations`, and a second
misleading warning (`"Solver took 1000 iterations."`) followed.

The user-visible cause is an `f_abstol` below the residual's own round-off floor — and
`f_reltol` does *not* rescue it, since it is anchored at `‖F(x₀)‖` and therefore becomes
*tighter* the better the initial guess is. That trap is now documented in `Options`.

## Changes

**1. Algorithm.** An explicit round-off resolution `τ = τ_ulps·ulp(φ(0))`, from which the
smallest *informative* step `αmin = τ/(c₁|φ'(0)|) = 2·τ_ulps·α*` follows, so the search stays
clear of the region where the test is decided by rounding. The condition is
`φ(α) ≤ min{φ(0), φ(0) + c₁αφ'(0) + τ}` — the `min` bounds the slackening, so `τ` may reduce
the decrease *demanded* but can never accept a merit above `φ(0)`.

`τ_ulps` is **precision-aware** (`armijo_ulps(T, c₁)`). `τ` must be at least ~an ulp of `φ(0)`
to recognise a merit at its round-off floor, and far below the `2c₁·φ(0)` the condition demands
at `α = 1` to leave that condition meaningful — compatible only while `eps(T) ≪ 2c₁`. That holds
by ten orders of magnitude in `Float64` and by a factor 400 in `Float32`, so the nominal 4 ulps
stands in both and their behaviour is *bit-identical* to a fixed 4 (verified over five methods ×
four fixtures). It fails outright in `Float16`, where `eps(T) = 9.8e-4` already exceeds
`2c₁ = 2e-4`: the nominal `τ` was twenty times the demanded decrease, which both removed the
sufficient-decrease safeguard at every `α` and classified a genuine two-ulp decrease as
`LINESEARCH_FLOOR` — which `solver_step!` counts towards `max_stalls`, so a converging
half-precision solve could be reported as stagnated. The cap (at one percent of the demanded
decrease) drops `Float16` to ~2e-3 ulps, in effect the exact condition; the floor is still
detected without `τ`, both by the condition degenerating to `φ(α) ≤ φ(0)` at a small enough step
and by the bit-frozen-merit check.

Plus an up-front descent check (a non-descent anchor no longer costs 53 wasted merit evaluations), bit-frozen-merit
detection, and safeguarded quadratic/cubic interpolation confined to `[0.1α, p·α]` — so `p` is
now an *upper bound* on the shrink factor and the trial sequence is pointwise never longer
than before.

**2. Public status API.** `solve_with_status` returns a `LinesearchStatus` carrying
`(α, outcome, trials, φ₀, d₀, φ, τ, αmin)`, with a `LinesearchOutcome` enum that separates a
genuine decrease from an irreducible merit — the distinction the step length alone cannot
express. All six methods report a genuine outcome; `solve` keeps returning `T`, so no caller
breaks.

**3. Plumbing.** `config(linesearch(s)) === config(s)` now holds for `NewtonSolver`,
`PicardSolver` and `DogLegSolver`, so `verbosity = 0` finally silences the line search
(downstream had to use a `NullLogger`), and `max_iterations` no longer doubles as the ladder
budget — that is now `linesearch_max_iterations`, defaulting to `linesearch_iterations(T)`
(60 for `Float64`) rather than 1000. Both quadratic searches use the same field, so their
budget is settable too.

**4. Termination, and acting on what the line search reports.** A stagnating solve stops after
`max_stalls = 2` steps instead of running to `max_iterations`, with one actionable warning
naming the achieved residual against the requested tolerance. A stall is diagnosed from the
step actually taken, so it also covers a `Static` step along an underflowed direction, a
collapsed DogLeg radius and an expanding Picard map. A stall now also **forces a fresh Jacobian
on the next step** (`maybe_refactorize!(…; stalled)`), which is what makes `max_stalls = 2`
conclusive for any `refactorize` rather than only for `refactorize = 1`. And a
`LINESEARCH_NO_DESCENT` report is acted on: the iterate is left in place — moving along a
direction the line search rejected outright would only make the retry start from a worse point
— and a stall recorded, which triggers exactly the Jacobian refresh `check_anchor` prescribes.
`solve!` also tests the stopping criteria *before* the first step, and
`SimpleSolvers.status(solver, state)` makes the outcome inspectable (`solve!` returns `x`, so
it was previously unreachable).

**5. The line-search contract** (`638e956`). Investigating a crash showed the six line searches
differing along eight independent axes, only some principled. Every method reached through
`solve`/`solve_with_status` now guarantees: it never throws (unhandleable situations are
reported, not raised); it returns `α > 0`; it reports through `linesearch_warnings` only; a
non-finite or ascending anchor is reported rather than assumed away; and its cost is bounded
independently of the merit's scale. What is deliberately *not* standardised is the meaning of
the input `α`, because there are two families — condition-satisfying and `α`-relative
(`Backtracking`, `StrongWolfe`, `Static`) versus minimising and `α`-independent (`Bisection`,
`Quadratic`, `BierlaireQuadratic`, which check no Wolfe condition).

## Fixed along the way

- **`BierlaireQuadratic` could abort a whole solve** with `ERROR: The function f must be
  decreasing at 0.0` from `triple_point_finder` — 3 of 300 random starts. Two causes, both
  measured: an ascent anchor (`φ'(0) = +2.06e-15` for `Float64`, `+1.10e-02` for `Float32`,
  both at `refactorize = 5`), and a merit flat to round-off. Halving `δ` answers neither.
- **`BierlaireQuadratic` could stall on a single point** until its budget ran out: `φ(α) =
  c(α−1)²` cost 15/70/14 evaluations at `c = 1/1e-6/1e-12`, evaluating
  `α = 0.99999999999999911` thirty-plus times. When `χ` lands on `b` the bracket never
  contracts. Bisecting the wider sub-interval fixes it; cost is now 15/16/16/15/15 across merit
  scales `1e-12 … 1e12`. Over 300 random starts the `Float64` 95th-percentile distance to the
  root improves from 9.7e3 to 0.5 `eps` with a fresh Jacobian and from 8.5e4 to 1.6 `eps` with
  `refactorize = 5` (median 8599 → 0.50).
- `StrongWolfe` threw an `AssertionError` on a NaN merit or derivative, and could return
  `α = 0` from its zoom phase. It also evaluated the merit **twice per trial step** (once
  directly, once inside the one-argument `sdc`) plus once more when building its status; all
  three now reuse the value in scope.
- `Quadratic` and `Bisection` could return negative steps — up to `α = -3`, 49 of ~3750 calls
  at `refactorize = 5` — inherited from `bracket_minimum`'s direction flip. Verified: 0 negative
  steps in ~45000 calls.
- `triple_point_finder` conflated two opposite failures under one `nothing`. It now returns
  `:flat` (no resolvable decrease → `FLOOR`) or `:unbracketable` (a decrease that could not be
  bracketed → `EXHAUSTED`); conflating them made the outer iteration count a *descending* merit
  as stagnation.
- `LinesearchStatus.trials` was a hardcoded `0` for the three minimising searches, so the floor
  message read "in 0 trial step(s)". It is a real count everywhere now, exact for `Backtracking`
  and `StrongWolfe`.
- The stagnation warning had no `maxlog`, so a caller looping `solve!` per time step would see
  it once per step — reproducing the flood this PR removes.
- The round-off resolution was a fixed 4 ulps at every precision, which in `Float16` exceeds the
  decrease the sufficient decrease condition demands (see change 1 above).

## Verification

- **4116 assertions pass**; Aqua clean; docs build clean (doctests and cross-references). The
  contract matrix (6 methods × 8 pathological anchors) and the scale-invariance test run for
  `Float64`, `Float32` *and* `Float16` — the precision where the `τ` cap and `αmin`'s √eps clamp
  are load-bearing rather than decorative.
- **`Float64`/`Float32` are bit-identical** to the pre-`armijo_ulps` state: every returned step,
  outcome, trial count, `τ` and `αmin` matches across five methods × four fixtures × two
  precisions. The cap only ever binds in `Float16`.
- **Measured against SimpleSolvers 0.9.2** in a throwaway environment over 15
  GeometricProblems integrations: **402 → 4** user-visible messages. The remaining 4 are the
  genuinely unsolvable case (`ThreeBody` with a collisional initial condition, where no root
  exists), down from 402 for that case alone. Every other integration went to **0**. 12 of 14
  trajectories bit-for-bit identical; the exceptions are Lyapunov amplification of round-off in
  `LotkaVolterra2d` lode Gauss(2) (29 eps by step 1000) and the collisional `ThreeBody`.
- Throw sweep (300 starts × 5 methods × 2 precisions × 2 `refactorize` settings): no
  "must be decreasing", no "Unable to bracket", no `AssertionError`, no `α ≤ 0`.

⚠️ The downstream sweep predates the final commit. Freezing the iterate on
`LINESEARCH_NO_DESCENT` is the change that could move a trajectory, so it should be repeated
before merging.

One measurement changed the design: gating the round-off-floor message at `verbosity ≥ 1`
newly surfaced a warning on three *previously silent, correctly converging* solves. Reaching
the merit's floor is the normal final state of a converged solve, so that outcome (and the
stationary-anchor one) is reported only at `verbosity ≥ 2`; whether an irreducible merit
matters is the outer iteration's call, and it makes it.

## Breaking — version 0.10.0

`bracket_minimum` is **exported** and no longer throws, so `a, b = bracket_minimum(f, x)` on an
unbracketable `f` now fails with `MethodError: iterate(::Nothing)` where it used to raise a
descriptive error, and code relying on the throw to detect failure proceeds silently. With the
rest of the list below, a downstream `SimpleSolvers = "0.9"` bound must not pick this up
silently — hence the minor bump rather than 0.9.3.

- `assess_convergence` returns a 4-tuple `(x_converged, f_converged, f_increased, stalled)`.
  Existing `a, b, _ = …` destructurings keep working.
- `NonlinearSolverStatus` gained `stalls::Int`/`stalled::Bool`; `NonlinearSolverState` gained
  `stalls::Int`/`stallflag::Bool`; `Options` gained `linesearch_max_iterations`/`max_stalls`;
  `Backtracking` gained `τ_ulps::T` (positionally defaulted, and capped at `armijo_ulps(T, c₁)`
  by the inner constructor — so a `Float16` method gets ~2e-3 ulps whichever way it is built,
  including via `change_precision`).
- Removed: `max_number_of_quadratic_linesearch_iterations` and its two constants; the
  `solve(::Linesearch{<:Bisection}, α₀, α₁, params)` and `BierlaireQuadratic`
  `solve(ls, α₀, params, iteration_number)` overloads.
- `triple_point_finder` returns `Union{Symbol,…}` on failure (unexported).
- Exported: `LinesearchStatus`, `LinesearchOutcome`, the six `LINESEARCH_*` values,
  `solve_with_status`, `NonlinearSolverStatus`. Deliberately **not** exported:
  `issufficient`, `isfloor`, `steplength`, `outcome`, `trials`, `isconverged`, `isstalled`,
  `status` — generic names a package doing `using SimpleSolvers` may want for itself, `status`
  most of all (a downstream definition would be a method-definition error, not a shadowing
  warning). Reach them as `SimpleSolvers.status` and so on.
- With `min_iterations = 0` and `f_abstol > 0`, an already-satisfying initial guess now takes
  **zero** steps instead of one, and `f_abstol_break` can fire before any step.

Three existing tests were re-recorded, all for substantive reasons: the `φ'(0) > 0` exhaustion
test (now caught up front), the NaN-detection test (`exp(-1/x²)` is exactly `0` at `x = 0`, so
the pre-step check correctly returns without stepping from a root; it now passes
`min_iterations = 1`), and the `τ` slackening test (the `min` against `φ(0)` correctly rejects
what the unbounded form accepted).

Also adds the 0.9.1/0.9.2 CHANGELOG sections that were never written, and marks the 0.9.0
"default line search is now `StrongWolfe`" entry as reverted in 0.9.1 (commit 264be0a).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,9 +64,24 @@ better the initial guess is.
   caller already computed, so neither `Backtracking` nor `StrongWolfe` evaluates the merit twice
   per trial. `StrongWolfe` keeps `τ = 0`. The condition is
   `fα ≤ min(f₀, f₀ + cαd₀ + τ)`: the allowance may reduce the decrease *demanded*, but it never
-  accepts a step whose merit exceeds `f₀`. The `min` is inactive in double precision and
-  necessary in `Float16`, where four ulps of `f₀` is `3.9e-3` — twenty times the `2c₁ = 2e-4`
-  demanded at `α = 1`, so an unbounded `τ` would license a visible uphill step.
+  accepts a step whose merit exceeds `f₀`.
+- `armijo_ulps(T[, c₁])` makes the round-off resolution **precision-aware**, and is what every
+  method now uses in place of the bare `DEFAULT_ARMIJO_τ_ULPS`. `τ` must be at least ~an ulp of
+  `φ(0)` to recognise a merit at its round-off floor, and far below the `2c₁·φ(0)` the condition
+  demands at `α = 1` to leave that condition meaningful; the two are compatible only while
+  `eps(T) ≪ 2c₁`. That holds by ~10 orders of magnitude in `Float64` and by a factor 400 in
+  `Float32`, so the nominal 4 ulps stands in both and their behaviour is bit-identical to before.
+  It fails outright in `Float16`, where `eps(T) = 9.8e-4` already exceeds `2c₁ = 2e-4`: the
+  nominal `τ` was *twenty times* the demanded decrease, which degenerated the Armijo test to
+  plain monotonicity at every `α` and — worse — classified a genuine two-ulp decrease as
+  `LINESEARCH_FLOOR`, which `solver_step!` feeds to `flag_stall!`, so a *converging* half-precision
+  solve could be reported as stagnated. The cap (`ARMIJO_τ_DEMAND_FRACTION = 0.01`, i.e. `τ` may
+  distort the demanded decrease by at most one percent) drops `Float16` to ~2e-3 ulps, in effect
+  the exact condition. Nothing is lost: the floor is still detected without `τ`, both by the
+  condition degenerating to `φ(α) ≤ φ(0)` at a small enough trial step and by `Backtracking`'s
+  two-consecutive-bit-identical-merits check. `Backtracking` applies the cap in its inner
+  constructor, so every path in — including `change_precision` and an explicitly oversized
+  `τ_ulps` — gets a resolution the element type can support.
 - `Options` gained `linesearch_max_iterations` (default `linesearch_iterations(T)`: 60 for
   `Float64`, 31 for `Float32`, 18 for `Float16`) and `max_stalls` (default `MAX_STALLS = 2`).
 - `SimpleSolvers.with_config(ls, config)` returns a `Linesearch` with the problem and method of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to SimpleSolvers.jl are documented here.
 
-## [0.9.3]
+## [0.10.0]
 
 Robustness release for the `Backtracking` line search. Downstream packages
 (GeometricIntegrators, GeometricProblems, ChargedParticleDynamics) were flooded by
@@ -40,33 +40,45 @@ better the initial guess is.
 
 - `LinesearchStatus`, `LinesearchOutcome` (`LINESEARCH_DECREASED`, `LINESEARCH_FLOOR`,
   `LINESEARCH_EXHAUSTED`, `LINESEARCH_NO_DESCENT`, `LINESEARCH_STATIONARY`,
-  `LINESEARCH_UNKNOWN`), `solve_with_status`, `issufficient` and `isfloor`. `solve_with_status`
+  `LINESEARCH_UNKNOWN`) and `solve_with_status` are exported; the accompanying predicates and
+  accessors `issufficient`, `isfloor`, `steplength`, `outcome` and `trials` are *not*, since
+  they are generic names a package doing `using SimpleSolvers` may want for itself — reach them
+  as `SimpleSolvers.steplength` and so on. `solve_with_status`
   returns the step length *plus* why the search stopped, which the step length alone cannot
   express — a tiny `α` may be the right answer or all that is left after the merit turned out
   to be irreducible. All six built-in methods report a genuine outcome (see the contract section
   below); the generic `LINESEARCH_UNKNOWN` fallback remains for user-defined methods.
   `SimpleSolvers.linesearch_warnings` is now the single place a line search emits messages.
-- `Backtracking` gained a `τ_ulps` keyword (default 4): the round-off allowance
-  `τ = τ_ulps·ulp(φ(0))` added to the sufficient decrease condition, from which the smallest
-  *informative* step `αmin = τ/(c₁|φ'(0)|) = 2·τ_ulps·α*` is derived. Because `αmin` is a
-  factor `2·τ_ulps` above `α*`, the search provably never enters the region where the test is
-  decided by rounding. `τ_ulps = 0` recovers the exact condition. New (unexported)
+- `Backtracking` gained a `τ_ulps` keyword (default 4): the round-off *resolution* of the merit,
+  `τ = τ_ulps·ulp(φ(0))`, from which the smallest *informative* step
+  `αmin = τ/(c₁|φ'(0)|) = 2·τ_ulps·α*` is derived. Because `αmin` is a factor `2·τ_ulps` above
+  `α*`, the search does not enter the region where the test is decided by rounding — unless
+  `backtracking_αmin`'s upper clamp at `√eps(T)` binds, which it does for a very flat merit in
+  double precision and for essentially any merit in `Float16`. Trial steps below `α*` are safe
+  there because the condition reduces to plain monotonicity (see below), and such an accept is
+  classified `LINESEARCH_FLOOR`. `τ_ulps = 0` recovers the exact condition. New (unexported)
   `armijo_tolerance`, `backtracking_αmin`, `backtracking_interpolation`, and the constants
   `DEFAULT_ARMIJO_τ_ULPS` and `BACKTRACKING_SHRINK_MIN`.
 - `SufficientDecreaseCondition` gained a keyword-only `τ` field (default `zero(T)`, i.e. the
   exact former condition) and a two-argument call form `sdc(α, fα)` taking a merit value the
-  caller already computed, so a backtracking loop no longer evaluates the merit twice per
-  trial. `StrongWolfe` keeps `τ = 0`.
+  caller already computed, so neither `Backtracking` nor `StrongWolfe` evaluates the merit twice
+  per trial. `StrongWolfe` keeps `τ = 0`. The condition is
+  `fα ≤ min(f₀, f₀ + cαd₀ + τ)`: the allowance may reduce the decrease *demanded*, but it never
+  accepts a step whose merit exceeds `f₀`. The `min` is inactive in double precision and
+  necessary in `Float16`, where four ulps of `f₀` is `3.9e-3` — twenty times the `2c₁ = 2e-4`
+  demanded at `α = 1`, so an unbounded `τ` would license a visible uphill step.
 - `Options` gained `linesearch_max_iterations` (default `linesearch_iterations(T)`: 60 for
   `Float64`, 31 for `Float32`, 18 for `Float16`) and `max_stalls` (default `MAX_STALLS = 2`).
 - `SimpleSolvers.with_config(ls, config)` returns a `Linesearch` with the problem and method of
   `ls` but different `Options`.
-- `NonlinearSolverStatus`, `isconverged`, `isstalled` and `status(solver, state)` are exported.
-  `solve!` returns `x`, not a status, so `status(solver, state)` is how a caller inspects the
-  outcome — in particular whether a solve *converged* or merely *stagnated* at the residual
-  floor. (`status` was removed from the exports in 0.9.0 as undefined; this implements it.)
+- `NonlinearSolverStatus` is exported; `SimpleSolvers.isconverged`, `isstalled` and
+  `status(solver, state)` are not, `status` least of all — a downstream package that does
+  `using SimpleSolvers` and defines its own `status` would get a method-definition error rather
+  than a shadowing warning. `solve!` returns `x`, not a status, so
+  `SimpleSolvers.status(solver, state)` is how a caller inspects the outcome — in particular
+  whether a solve *converged* or merely *stagnated* at the residual floor.
 - `SimpleSolvers.residual_small`, `iterate_settled`, `stalled_step`, `record_stall!`,
-  `flag_stall!` and `stall_number`.
+  `flag_stall!`, `stall_number` and `needs_refresh`.
 
 ### Changed (behavioural)
 
@@ -113,11 +125,27 @@ better the initial guess is.
   and convergence are mutually exclusive by construction. This is diagnosed from the step
   actually taken, so it also covers a `Static` step along an underflowed direction, a collapsed
   `DogLeg` trust-region radius and a locally expanding `Picard` map; a line search that reports
-  `LINESEARCH_FLOOR` additionally flags it one iteration earlier. A counter (rather than a
-  one-shot stop) is used because `maybe_refactorize!` and `DogLegSolver` both deliberately
-  recover on the step *after* a frozen one. `max_stalls = typemax(Int)` restores the old
-  behaviour. The two misleading warnings are replaced by one message naming the achieved
-  residual, the requested tolerance and both remedies.
+  `LINESEARCH_FLOOR` or `LINESEARCH_NO_DESCENT` additionally flags it one iteration earlier.
+  A counter (rather than a one-shot stop) is used because the step *after* a stalled one is
+  deliberately attempted under better conditions — see the next entry. `max_stalls = typemax(Int)`
+  restores the old behaviour. The two misleading warnings are replaced by one message naming the
+  achieved residual, the requested tolerance and both remedies, rate-limited with `maxlog` so
+  that a caller looping `solve!` per time step does not see it once per step.
+- **A stalled step forces a fresh Jacobian on the next step.** `maybe_refactorize!` gained a
+  `stalled` keyword, supplied by `solver_step!` from the new `SimpleSolvers.needs_refresh(state)`.
+  Without it, a quasi-Newton solver with `refactorize = r` would rebuild the same direction from
+  the same stale Jacobian for up to `r - 1` further steps and reproduce the same negligible step,
+  so `max_stalls = 2` could give up at the second of those for a reason a fresh Jacobian would
+  have fixed. Now the second consecutive stall is one a freshly evaluated Jacobian did *not* fix,
+  which makes `max_stalls = 2` conclusive for every `refactorize` rather than only for
+  `refactorize = 1`. `DogLegSolver` folds the same signal into its existing `force_refresh`.
+- **`LINESEARCH_NO_DESCENT` is acted on rather than only reported.** An ascent anchor is a
+  stale-Jacobian symptom whose remedy is to refresh the Jacobian, so `solver_step!` now leaves the
+  iterate where it is — moving along a direction the line search has rejected outright would only
+  make the retry start from a worse point — and records a stall, which triggers exactly that
+  refresh. Previously the full step was taken along the ascent direction, the iterate moved (so
+  nothing counted it as a stall), and the solve ran to `max_iterations`. The line search itself
+  still returns `α > 0` as its contract requires; whether to use it is the caller's decision.
 - **`solve!` tests the stopping criteria before the first step.** An initial guess that already
   satisfies them is no longer perturbed by a full solver step (Jacobian, factorization, and a
   line search asked to improve an already-exact residual). Only the absolute branch is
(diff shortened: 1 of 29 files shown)


michakraus (commented):
# Review: PR #171 — robust backtracking + standardised line-search contract

Reviewed all three commits (`c371d10`, `0a315b3`, `638e956`), +2271/−187 / 29 files. CI green on the new head.

## Overview

The core diagnosis is right and is the most valuable part: `eps(1.0)` in the reported `α` proves the ladder always exited on its `α ≤ eps` guard, so the "within 1000 iterations" was never true. The follow-up commits are a bigger win than the original: `638e956`'s five-point contract on `LinesearchMethod` (never throws, always returns `α > 0`, reports only through `linesearch_warnings`, validates the anchor, cost independent of merit scale) turns six divergent implementations into one, and the two-family distinction (condition-satisfying/α-relative vs. minimising/α-independent) is the right seam to leave un-standardised.

Verified while reading:

- `backtracking_interpolation`'s quadratic and cubic models are correct — Cramer determinant `α²αp²(α−αp)`, coefficients, minimiser `(−b+√(b²−3ad₀))/(3a)`, degenerate `a = 0` branch.
- The ladder is provably strictly decreasing (`αₙ ≤ pα < α`, `αmin < α`), so termination does not rely on the budget.
- The `params.φ₀` short-circuit is safe: `nan_recovery!` writes trial iterates to `solution(cache(s))`, never to `x`, so `value(state)` still matches `params.x`.
- `record_stall!` runs exactly once per iteration and after `update!`, so the counter is current at the top-of-loop test; `stalled` and `isconverged` are mutually exclusive by construction (both gate on `residual_small`).
- The iteration-0 stopping test can only fire on the `f_abstol`/`f_abstol_break` branches, as documented.
- The `BierlaireQuadratic` stall analysis is correct: with `χ == b` both branch tests are false, `c ← b` collapses the triple, and the `(a+c)/2` fallback plus the shift map back onto `b`. Bisecting the wider sub-interval does make `c − a` contract strictly.

Test coverage is genuinely strong, and the contract matrix (6 methods × 8 pathological anchors, asserting no-throw and `α > 0`) is exactly the right shape.

## Findings

### 1. `check_anchor`'s documented remedy is never taken; the ascent step is taken at full length

`check_anchor`'s docstring says of an ascent anchor: *"The correct response is to refresh the Jacobian (see `maybe_refactorize!`), which is why the line search reports the situation instead of trying to salvage a step from it."* But nothing acts on the report. `solver_step!` inspects only `isfloor(lsstatus)`; `LINESEARCH_NO_DESCENT` is discarded, `x` moves by the **full** returned step along an ascent direction, `allow_f_increases` defaults to `true`, and because the iterate *did* move, `iterate_settled` is false — so no stall is recorded and the solve still spins to `max_iterations`.

This is not hypothetical: `638e956`'s own message reports measured ascent anchors (`φ'(0) = +2.06e-15` for Float64, `+1.10e-02` for Float32, both at `refactorize = 5`). Suggest making the remedy real — a `flag_refresh!` on the state that makes the next `maybe_refactorize!` `force`, mirroring `DogLegSolver`'s `force_refresh` — and/or `flag_stall!` so the stagnation path produces the actionable message. As it stands the one failure mode the PR doesn't fix is also the one most likely to diverge rather than freeze.

### 2. `τ = τ_ulps·ulp(φ₀)` is not scale-free, and `Float16` decreases get classified as `FLOOR`

`τ/φ₀` is `8.9e-16` in Float64 and `4.8e-7` in Float32 — comfortably below the `2c₁ = 2e-4` the Armijo condition demands at `α ≈ 1`, exactly as `DEFAULT_ARMIJO_τ_ULPS`' docstring argues. In **Float16** it is `4·9.77e-4 = 3.9e-3`, i.e. **20× larger** than the demanded decrease.

Two consequences, and `638e956` widened the second one from `Backtracking` to all six methods:

- `Backtracking(Float16)` accepts `α = α₀` whenever `φ(1) ≤ 1.0039·φ₀`, i.e. it licenses a merit *increase* of up to 0.4%.
- Every method now classifies its result with `φres ≤ φ₀ - τ ? LINESEARCH_DECREASED : LINESEARCH_FLOOR`. In Float16 a genuine 0.2% decrease is therefore labelled `FLOOR`, which `solver_step!` turns into `flag_stall!` — two such steps and the solve aborts as "stagnated" while it was in fact converging.

Float16 is a supported precision here (`change_precision`, `linesearch_iterations(Float16) == 18`, and `14bf15a` was a Float16 fix), but the contract matrix and the scale-invariance test are Float64-only, so nothing catches this. Suggest making the default `τ_ulps` a function of `T`, or capping `τ` at something like `c₁·√eps(T)·|d₀|` — which is what the `αmin` clamp already implies.

### 3. The "provably never enters the rounding region" guarantee holds only when the `√eps(T)` clamp is inactive

`αmin_raw = 8α*`, but `backtracking_αmin` caps `αmin` at `√eps(T)`, so `αmin < α*` whenever `|d₀| < ulp(φ₀)/(2c₁√eps(T))` — for Float32 with `φ₀ = 1` that is `|d₀| < 1.7`, which includes the exact-Newton `d₀ = −2φ₀` for `φ₀ ≲ 0.85`; for Float16 it is essentially always. `Backtracking`'s docstring, `backtracking_αmin`'s docstring and `docs/src/linesearch/backtracking.md` ("the search never enters that region at all") all state the factor-`2n` result unconditionally. Classification stays correct (`FLOOR`, not a bogus `DECREASED`), so this is documentation accuracy — worth one qualifying clause in each of the three places.

### 4. `max_stalls = 2` can abort a `refactorize ≥ 4` quasi-Newton solve

`MAX_STALLS`' docstring argues two is conclusive because "a quasi-Newton solver refreshes its `Jacobian` on a later step". But `maybe_refactorize!` refreshes only at `mod(iteration, refactorize) == 0 || iteration ≤ 1`, so with `refactorize = 5` iterations 6–9 all run on the stale Jacobian and two consecutive frozen steps abort at iteration 7 instead of recovering at 10. (The `DogLegSolver` half of the same claim *is* accurate — `solver_step!` resets a collapsed `Δ` and forces a refresh on the next step.) Default `refactorize = 1` is unaffected. Same fix as finding 1: force a refactorization when `stall_number(state) > 0`. Failing that, correct the docstring and consider defaulting `max_stalls` to `max(MAX_STALLS, refactorize)`.

### 5. The new stagnation warning has no `maxlog`

Every *line-search* message here is rate-limited (`maxlog = 1`/`3`), but `nonlinear_solver_warnings`' stagnation warning is not — it fires once per `solve!`, so an integrator that calls `solve!` per timestep on a stagnating problem reproduces the flood this PR exists to remove, just with a better message. Conversely, `maxlog` is keyed on source location and therefore **process-global**: once three `EXHAUSTED` messages have been emitted, every later solve in the session is silent, including a genuinely new problem. Both look deliberate; neither is stated. Suggest a `maxlog` on the stagnation warning and a sentence in `linesearch_warnings`' docstring on the scoping.

### 6. Version should be `0.10.0`, not `0.9.3`

`638e956` added public breakage on top of the first commit's list. `bracket_minimum` **is exported** (`src/SimpleSolvers.jl:79`) and no longer throws: an external `a, b = bracket_minimum(f, x)` now fails with `MethodError: iterate(::Nothing)` instead of a descriptive error, and anyone who relied on the throw to detect failure silently proceeds. Also removed: `MAX_NUMBER_OF_ITERATIONS_FOR_QUADRATIC_LINESEARCH`, its `_SINGLE_PRECISION` sibling, `max_number_of_quadratic_linesearch_iterations`, the `solve(::Linesearch{<:Bisection}, α₀, α₁, params)` overload and the `BierlaireQuadratic` `solve(ls, α₀, params, iteration_number)` overloads. Plus the first commit's field additions to four structs and the tightened `LiSeT<:Linesearch{T}` constraint. A downstream `SimpleSolvers = "0.9"` bound picks all of this up automatically.

### 7. `trials` and `αmin` are placeholders in five of six methods

`LinesearchStatus.trials` is documented as "the number of trial steps `α > 0` at which the merit was actually evaluated", but `Bisection`, `Quadratic` and `BierlaireQuadratic` all hardcode `0`, and `StrongWolfe` counts only its expansion loop (not the zoom phase, not the extra `φ(αres)`). So the `verbosity ≥ 2` floor message reads "in 0 trial step(s)". Likewise `αmin = zero(T)` in those statuses makes `linesearch_warnings`' `EXHAUSTED` branch always select the "budget was spent" wording — correct for `Bisection` by accident, wrong in general. Either thread the real counts through (the loops already have them) or document the fields as `Backtracking`-only and skip them in the message.

### 8. `wolfe_status` re-evaluates the merit at the accepted step

Every `StrongWolfe` return path calls `φ(αres)` afresh — a full `F` evaluation per solver step for a `NonlinearSolver`, which partly undoes the `params.φ₀` saving from the first commit — and `n` is not incremented for it, so `trials` undercounts. In the `cc(αi) && return wolfe_status(αi)` path `φi` is already in scope, and `_wolfe_zoom` knows the merit at the point it returns. Threading the value through (as `Backtracking` does with the two-argument `sdc`) removes the cost.

### 9. `nothing` from the bracketers conflates two opposite situations

`triple_point_finder` returns `nothing` either because `fx₁ ≤ fx₀ + τ` (flat to round-off — `FLOOR` is right) or because `nmax` doublings never reached a turning point, i.e. the merit is still *descending*. `Bisection`/`Quadratic` map `bracket_minimum`'s `nothing` the same way. In the second case `FLOOR` is the wrong label and, via `flag_stall!`, actively counts a descending merit as stagnation. With `nmax = 100`, `k = 2`, `s = 1e-2` the bracket reaches `α ≈ 1e28`, so this is practically unreachable for a merit bounded below — but the code and docstrings read as though `nothing` had one meaning. A distinguishing return value, or a comment saying why the conflation is safe, would settle it.

### 10. The round-off allowance is duplicated in `triple_point_finder`

`τ = 4 * eps(fx₀)` is hardcoded with a comment noting that `armijo_tolerance`/`DEFAULT_ARMIJO_τ_ULPS` are unreachable given the `bracketing` → `linesearch` include order. `base/options.jl` is already included at line 49, well before `bracketing/` at line 81, so moving those two definitions there (or into a small `base/roundoff.jl`) removes the duplicate — and makes the Float16 fix from finding 2 a single-site change.

### Minor

- `capped = α > αmin` in `Backtracking`'s `solve_with_status` mislabels a budget exhaustion that lands exactly on `αmin`; a flag set where the `for` falls through would be exact. Message-only.
- `Bisection` and `Quadratic` return `LINESEARCH_NO_DESCENT` when the *result* is non-positive, though `check_anchor` has already established the anchor descends — `FLOOR` (or a new outcome) would describe it better.
- Export set is inconsistent: `LinesearchStatus`, `solve_with_status`, `issufficient`, `isfloor` are exported but the primary accessors `steplength`, `outcome`, `trials` are not (the docs example has to import them), and `τ`/`αmin` have no accessors at all, so tests and docs reach into fields. Separately, exporting the bare name `status` risks a method-definition error (not a shadowing warning) for any downstream package that does `using SimpleSolvers` and defines its own; `GeometricIntegrators` doesn't today.
- `linesearch_iterations` is derived for `p = 0.5`; a user-supplied `p` near 1 needs many more trials in principle, though the interpolation clamp keeps the frozen case near halving regardless. One sentence would cover it.
- `armijo_tolerance(φ₀::T, n::T)` needlessly forces the ulp count to the merit's element type; `Quadratic`'s `d₀anchor` breaks the naming convention of its neighbours.
- The inner `NonlinearSolver` constructor still accepts and silently discards `options_kwargs...` — pre-existing, but now dead weight since `config` is passed positionally.

### Coverage gaps

Matching the findings: no Float16 or Float32 run of the contract matrix or the scale-invariance test (findings 2 and 3 live there); no full-solve test of the `d₀ > 0` path (finding 1); no `refactorize ≥ 4` stall test (finding 4).

## Recommendation

Approve after findings 1, 5 and 6; findings 2 and 4 want either a fix or an explicit "won't fix, here's why"; 3, 7, 9 and 10 are cheap. The docs-only subset is 3, 4 (the `MAX_STALLS` claim), 5 (the `maxlog` scoping) and 9.

PR #179, opened 2026-08-13
Detect an overflowed residual, not just an undefined one

### LINKED ISSUE #130: Test NaN handling in nonlinear solvers

michakraus:
Closes #130.

## The gap

Every guard the nonlinear solvers had against a trial iterate leaving the region where the problem is representable tested `isnan` — and `isnan` is false for `Inf`.

Issue #130 asks whether that detection works, on the example `f(x) = 5|x-2|³ - 1/|x|`. Its singularity produces `Inf` at `x = 0`, not `NaN`, so **for the issue's own fixture the answer was no**. Nor did anything downstream of the guards catch it: `rfₐ > f_abstol_break` cannot, because `f_abstol_break` defaults to `Inf` and `Inf > Inf` is false.

The line searches never had this gap — `check_anchor` has always tested `isfinite` — and neither did the `Optimizer` removed in 0.11.0, which tested `isnan(f) || isinf(f)` ([#138](https://github.com/JuliaGNI/SimpleSolvers.jl/pull/138), commit `0f22e16`). The solvers were the only ones left looking for half the problem.

Reproducer, before this PR — `F(x) = 1/x - 1` from `x₀ = 2`, where the Newton step is exactly `d = -2` and lands exactly on `x = 0`:

| | before | after |
|:--|:--|:--|
| `F(x) = 1/x - 1`, `x₀ = 2`, `Static()` | steps onto `x = 0`, residual `Inf`, next step throws on the `NaN` direction | damps once, lands on the root `x = 1` |
| `Inf` direction | passes the guard, cannot be damped, solve stalls silently | `NonlinearSolverException` |

## Direction and residual are not the same problem

That distinction is what the fix turns on.

A non-finite **direction is rejected**, as a `NaN` one always was. It cannot be recovered from: `nan_recovery!` damps by a *factor*, and `Inf * nan_factor` is `Inf`, so the loop would spend its whole budget reproducing the same trial iterate. Rejecting it first is also what makes the damping sound — past that guard the direction is finite, so every halving really does shorten the step. Previously such a step was neither rejected nor damped: the line search saw an infinite directional derivative, returned `LINESEARCH_NO_DESCENT`, and the solve made no progress without saying why.

A non-finite **trial residual is damped**, exactly as a `NaN` one was — `nan_recovery!`, `report_dogleg_nan` and the dogleg trust-region rejection now all test `isfinite`.

`havenan` becomes `SimpleSolvers.havenonfinite`. It is unexported and internal (nothing downstream references it), so there is no deprecation.

## What else was wrong

Three defects sit next to this one and are fixed with it. **The first is the serious one** — it is the difference between a solve that fails and a solve that lies:

- **`residual_small` could report *convergence* on an overflowed solve.** It guarded only `isnan(r₀)`, so an infinite `‖F(x₀)‖` made the relative term `f_reltol * Inf = Inf` and the gate vacuously true. A residual of `1e10` counted as small, and the solve claimed success wherever it happened to land. An overflowed initial residual is no more a usable reference scale than an uninitialized one, and is now treated the same way.
- **`iterate_settled` counted an infinite step as a frozen iterate**, since `Inf ≤ ‖x‖·x_suctol` holds once `‖x‖` has overflowed too. An iterate that jumped to infinity has neither converged nor stagnated — it has broken down, which is what `havenonfinite` is for.
- **`nan_recovery!` at `nan_max_iterations = 0` left its cache untouched**, so the post-condition its docstring advertises — that `solution(cache)` and `value(cache)` hold *this* step's trial and its residual — was false. The `PicardSolver` step documents that it reads that cache rather than re-evaluating `F`, so it tested a stale residual and committed a stale iterate from the **previous** solver step. One trial is now always evaluated, and the damping happens *before* the trial it is for rather than after the one that failed: the budget bounds the damping, not the evaluation, and `value(cache)` is always `F(x + direction(cache))`. Picard additionally refuses to commit a trial whose *residual* is not finite, as `DogLegSolver` already did on radius underflow.

## Behaviour

Unchanged on any solve that stays finite, and at the same cost — one pass over the residual either way. Where it changes, it changes in the direction of saying so: **a non-finite direction now raises `NonlinearSolverException` where it used to stall silently.** A caller driving `solve!` in a time-stepping loop will see that as an exception rather than as a step that quietly achieved nothing, which is the one visible regression risk here.

## Tests

The fixture is `F(x) = 1/x - 1` from `x₀ = 2`, chosen because the arithmetic is exact in floating point: `F = -0.5`, `J = -1/x² = -0.25`, so the Newton step is `d = -2` and `x + d = 0` exactly, and one halving gives `d = -1` — the root itself. `Static()` evaluates no merit at all, so `nan_recovery!` is the only thing standing between the solver and `x = 0`.

New testsets in `test/nonlinear_solver_tests.jl` cover: the `Inf` damping path (`Float64` and `Float32`); rejection of a non-finite direction; `nan_factor` and `nan_max_iterations`, which had **no test at all** before, neither their defaults nor a non-default value; the status-level `havenonfinite` / `meets_stopping_criteria` / warning trio, also previously untested; the Picard direction guard and commit guard; and the `residual_small` / `iterate_settled` finiteness gates. The verbosity gate of `report_nan_direction` is now pinned alongside the two `DogLeg` reporters that already were.

## Verification

- Full suite green — 2980 nonlinear-solver assertions.
- Doctests pass; docs build resolves the new `@ref`s with no new warnings.
- Downstream **bit-identical** before and after: GeometricIntegrators × GeometricProblems, `LotkaVolterra2d` over 1000 steps under five integrators, plus the collisional `ThreeBody` initial condition that has no root. Nothing downstream references `havenan` or matches on the changed message texts.

## Review round

A second commit fixes what reviewing the first one turned up — two guards that did not do what their comments said, and three assertions that passed for the wrong reason:

- **`nan_recovery!` damped outside its budget.** Damping after the failed trial rather than before the next one meant one more damping than evaluation: `nan_max_iterations = 0` still halved the direction once (so it behaved exactly like `1`, and the `1/x - 1` solve *converged on the root*), and on every exhaustion path `value(cache)` was the residual of a trial built from a direction one factor away from the one the cache holds — which is the pair the Picard step reads. Raised by Copilot for the budget-0 half.
- **The Picard commit guard tested the trial iterate**, but the failure it names is a non-finite *residual*, and there the iterate is finite by construction. So it passed and the poisoned trial was committed: with `F` finite on `x ≤ 1` and `Inf` beyond, the solve returned `x = 1.0000000021684043` with `F(x) = Inf`.
- The `Inf` row of the status testset left `rxₛ`/`rfₛ` at `NaN`, so the old `havenan` accepted it too; `!iterate_settled(Inf, …)` was already false with `‖x‖ = 1`; and the Picard fixture never produced a non-finite trial at all, making its assertion vacuous. Two further testsets now pin the recovery budget and the direction/residual pairing.

All six new assertions fail against the first commit and pass against the second.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -237,9 +237,11 @@ Three defects next to it are fixed with it, the first of which is the serious on
   its docstring advertises — that `solution(cache)` and `value(cache)` hold this step's trial and
   its residual — was false. The `PicardSolver` step documents that it reads that cache instead of
   re-evaluating `F`, so it tested a stale residual and committed a stale iterate from the
-  *previous* solver step. One trial is now always evaluated; the budget bounds the damping, not
-  the evaluation. The Picard step additionally refuses to commit a trial that is not finite, as
-  `DogLegSolver` already did on radius underflow.
+  *previous* solver step. One trial is now always evaluated and the damping happens before the
+  trial it is for, so the budget bounds the damping rather than the evaluation and the cached
+  residual always belongs to the direction the cache holds. The Picard step additionally refuses
+  to commit a trial whose *residual* is not finite, as `DogLegSolver` already did on radius
+  underflow.
 
 Behaviour on any solve that stays finite is unchanged, and the guards cost the same as before —
 one pass over the residual either way. Where it changes, it changes in the direction of saying
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,16 @@ All notable changes to SimpleSolvers.jl are documented here.
 
 ## [0.11.0]
 
-Three independent changes. **Breaking**: `Backtracking` loses its `α₀` key and its positional
+Four independent changes. **Breaking**: `Backtracking` loses its `α₀` key and its positional
 constructor, and gains an opt-in expansion phase. Additive: a `NonlinearProblem` can be solved
 without assembling a solver by hand, and a `NonlinearSolver` reports — and, on request, stops —
 a solve that is *not progressing*, the case
 [issue #173](https://github.com/JuliaGNI/SimpleSolvers.jl/issues/173) found next to the
-`max_stalls` machinery of 0.10.0.
+`max_stalls` machinery of 0.10.0. Fixed: the nonlinear solvers looked for `NaN` where they
+should have looked for any non-finite value, so an *overflowed* residual passed every guard
+they have —
+[issue #130](https://github.com/JuliaGNI/SimpleSolvers.jl/issues/130) — and could be reported
+as convergence.
 
 ### Convenience entry points for solving a `NonlinearProblem`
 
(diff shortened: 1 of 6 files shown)


michakraus (commented):
Verified independently: read the full diff, then ran the branch to check the claims it makes, and validated each proposed fix in a scratch copy. **Full suite green with all five fixes applied** — `Nonlinear Solvers | 2980 2980`.

## What holds up

- **The central widening is right and complete.** `grep -n "isnan\|isinf" src/` leaves only `backtracking.jl:288` — a *sentinel* test (`αp` is `NaN` on the first backtrack), not a guard — and three `@assert !isnan(...)` in the line-search conditions, which sit downstream of `check_anchor`'s `isfinite`. No `havenan` reference survives anywhere in the repo.
- **`havenonfinite` cannot turn a bad solve into a converged one**, as its docstring claims: `residual_small(Inf)` is `Inf ≤ f_abstol + f_reltol·r₀`, false for any finite `f_abstol`.
- **Four of the six new testsets genuinely discriminate** against `main` (checked by re-deriving the old predicates): the Newton `Inf` damping, the non-finite direction rejection, `nan_factor`, the `nan_max_iterations = 0` Picard sentinel — and the `residual_small` assertion (old: `1e10 ≤ f_reltol·Inf` → `true`).

## What does not

Five findings, in the inline comments: two behavioural, three tests that pass for the wrong reason. Copilot's comment is one of the two — confirmed, with a second half it did not mention.

**1. The Picard commit guard tests the wrong quantity** (`picard_solver.jl:159`). It guards the trial *iterate*, but the failure its own comment names is a non-finite *residual* — and in that failure the iterate is finite by construction (`x` is, and the direction is, past the guard 27 lines above). So the guard passes and the poisoned trial is committed:

```julia
Fwall(y, x, p) = (y .= [xi ≤ 1 ? -1.0e10 : oftype(xi, Inf) for xi in x])
x = [1.0]; s = PicardSolver(x, Fwall, similar(x); verbosity = 0); state = SolverState(s)
solve!(x, s, state)
# committed x = [1.0000000021684043],  F(x) = [Inf],  havenonfinite(status) = true
```

`F` is finite on `x ≤ 1` and `Inf` beyond it, and `‖F‖ = 1e10` is large enough that even the underflowed backtracking step `α‖d‖` still crosses the wall — so no trial has a finite residual and the last one is committed anyway. With the suggested guard the same fixture leaves `x == [1.0]` and reports `isstalled`, exactly the outcome the comment describes.

**2. `nan_recovery!` damps outside its budget** (`nonlinear_solver.jl:189`) — Copilot's finding, confirmed, and there is more to it. Measured on this branch with `F(x) = 1/x - 1`, `x₀ = 2`, `d = -2`:

| `nan_max_iterations` | direction on return | cached trial / residual |
|---:|---:|:---|
| 0 | **−1.0** (damped once) | 0.0 / `Inf` |
| 1 | −1.0 | **0.0 / `Inf`** — built from `d = −2` |
| 2 | −1.0 | 1.0 / 0.0 |

Row 0 is Copilot's point: a budget of zero damps once, so it behaves exactly like a budget of one — the `Finv` solve *converges on the root* at `nan_max_iterations = 0`. Row 1 is the second half: on **every** exhaustion path the direction has been damped one time more than the trial the cache holds, so `value(cache(s))` is not `F(x + direction(cache(s)))`. The Picard step relies on precisely that identity — "whose residual `‖F(x + d)‖` is already in `value(cache)` from the NaN loop above" (`picard_solver.jl:139`) — and then backtracks with an `α` measured against a mismatched direction.

Damping *before* the trial it is for fixes both at once, at the cost of at most one extra `F` evaluation on the exhaustion path. It also makes the CHANGELOG's "the budget bounds the damping, not the evaluation" literally true.

**3–5. Three assertions pass for the wrong reason** — details inline. Briefly: the `Inf` row of the status loop has `rxₛ = rfₛ = NaN` after a single `update!`, so the old `havenan` returns `true` there too (the comment claims the `Inf` rows "are the ones that used to fail" — they would not have); `!iterate_settled(Inf, …)` was already `false` with `‖x‖ = 1`, since the docstring's case needs `‖x‖` itself overflowed; and the `Fneg` fixture in the Picard testset never produces a non-finite trial at all — instrumented, the backtracking underflows `α` on the first step and the iterate freezes at `1.0000000000000009`, so `@test all(isfinite, x2)` is vacuous.

## Nit (no change suggested)

`assess_convergence` computes `f_settled = rfₛ ≤ norm(value(state)) * config.f_suctol` inline — the exact parallel of `iterate_settled`, left un-widened. Not reachable as a bug (`f_converged` gates it behind `residual_small`, which no infinite residual passes), but the asymmetry with the `iterate_settled` change is worth a sentence in the docstring.

## Verification

Suggestions 1 and 2, plus regression tests for both and the three strengthened assertions: `Nonlinear Solvers | 2980 2980`, everything else green. Run against *this branch's* source, the six new assertions fail 6/6 — so they pin the fixes rather than restate them.

PR #180, opened 2026-08-14
Tell a located root from a failed bracket, and report to the caller

michakraus:
Fixes the two upstream defects [GeometricOptimizers.jl](https://github.com/JuliaGNI/GeometricOptimizers.jl) filed against SimpleSolvers in its CHANGELOG (**D3** and **D4**), plus what reviewing that work turned up. Breaking: `NonlinearSolverStatus` gains a field, a `NonlinearSolver` no longer emits line-search warnings from inside its iteration, and a `LinesearchMethod` implements `solve_with_status` rather than `solve`. Version bumped to `0.12.0`.

## 1. `Bisection` reported success when it could not bracket (D3)

`_bisection_core` returned `converged = true` on the branch that handles *no sign change between the endpoints*:

```julia
if y₀ * y₁ > zero(y₀)                                  # no sign change
    report_bisection_nobracket(α₀, α₁, y₀, y₁, config)
    return (abs(y₀) ≤ abs(y₁) ? α₀ : α₁), true, n      # ← converged = true
end
```

The intent was that a flat derivative at a minimum is benign. The effect was that the core could not tell *"found the root"* from *"gave up"*, and the endpoint of a bracket that never existed was handed on as if it were the located line minimiser.

That claim then propagated. `Bisection` bisects φ′ but brackets on φ, so the two disagree exactly when something is wrong — a stale or regularized Jacobian, an inexact linear solve, a non-smooth merit. The returned endpoint then typically does not improve the merit either, and the search classified it as `LINESEARCH_FLOOR`: *"no line search can make progress here"*. The outer iteration acts on that (`flag_stall!`, then `max_stalls`), so a failure to bracket was being read as a property of the problem.

- `_bisection_core` returns a `BisectionOutcome` (`BISECTION_CONVERGED` / `BISECTION_NOBRACKET` / `BISECTION_EXHAUSTED`) in place of the `Bool`. The `α` it returns is unchanged in every case — only the claim about it is.
- The line search classifies a failed bracket **by the merit alone, with `LINESEARCH_FLOOR` disallowed**: `LINESEARCH_DECREASED` when the step still beats φ(0) by more than τ, `LINESEARCH_EXHAUSTED` when it does not. `LINESEARCH_FLOOR` is now reachable only from a bisection that actually converged. The retry from the α = 0 anchor carries its own verdict through for the same reason; it used to be discarded.
- `bisection` dispatches on the outcome, so a spent budget and a failed bracket get the wording that fits each — only the first is fixed by a larger budget. Its no-bracket warning moves to `verbosity ≥ 1`; it sat at 2 only because the line search used to route through `bisection`, which it no longer does.

## 2. `maxlog` was per session, not per solve (D4)

The caps were keyed on the *source location* of the `@warn`, so they were process-global and never reset between `solve!` calls — a genuine line-search failure in the tenth solve of a run was silent because the first had spent the budget. This mattered because the log was the only channel by which a rejected line search was visible.

**The caps were a symptom.** The channel split already existed — `solve_with_status` returns a `LinesearchStatus` and emits nothing, `solve` adds `linesearch_warnings` — but `solver_step!` called `solve_with_status` and then `linesearch_warnings` *anyway*, once per outer iteration. A solve that cannot make progress asks the line search for an impossible decrease at every one of its iterations, so that call is what turned one diagnosis into thousands of identical messages.

- `solver_step!` no longer calls `linesearch_warnings`. A line search reports to its **caller** through `LinesearchStatus`, and to the **user** only when the user called `solve(ls, α)` directly. One direct call yields one message, so the three `maxlog` caps are **deleted rather than reimplemented**.
- `NonlinearSolverState` tallies the outcome of every step; `NonlinearSolverStatus` carries a snapshot, read with `linesearch_outcomes` / `linesearch_failures` / `dominant_linesearch_outcome`. This is the programmatic channel the package lacked — the upstream half of GeometricOptimizers' open **B1**.
- The stagnation and no-progress messages name what the line search reported and how often, so a failed solve names the cause and not only the symptom.
- One deliberate loss: a solver at `verbosity ≥ 2` no longer pays for `curvature_diagnostic` (a full Jacobian) once per iteration.

`nonlinear_solver_warnings` is then the only site still needing a cross-solve cap. It uses `should_report!` — 1st, 2nd, 4th, 8th … occurrence — keyed on the dominant line-search outcome, so a solve that starts failing for a *new* reason is reported at once. The trade is stated rather than glossed: a repeating diagnosis is silent on occurrence 10 and reported on 16.

## Verification

- Full suite green: 1409 line-search and 3017 nonlinear-solver assertions (1398/3006 before the review commit). Docs build clean, doctests pass.
- Allocations unchanged: zero for every line search and for Picard/DogLeg (StrongWolfe's 288 B is the pre-existing documented exception).
- Downstream, in a throwaway environment: **ThreeBody/ImplicitMidpoint goes from 6 messages to 1**, and that one message now carries what the five suppressed line-search warnings used to say, with a count. LotkaVolterra2d/Gauss(2) is bit-identical.
- The GeometricIntegrators suite passes against this branch with Pass/Broken counts identical to registered 0.11.0.

## 3. Review pass (second commit)

A converged solve floors on its last step as a matter of course, and `linesearch_failures` counts `LINESEARCH_FLOOR`, so the new `verbosity ≥ 2` report announced that *every* healthy solve "did not go smoothly" — six iterations on `x² − 2`, one floor, one message. It now counts the outcomes that are **not** the expected floor, and names the one that made it fire rather than the floor it is ignoring (with one of each, the tie rule would otherwise explain the message by the half of it that is expected). `dominant_linesearch_outcome` gains a `count_floor` argument and `linesearch_reason` an outcome to explain.

Two of the **Open Issues** are closed rather than recorded:

- **A `LinesearchMethod` implements `solve_with_status`; `solve` is derived** (breaking). The layering held for the six built-ins only because each wrote the same three-line `solve` on top of its status; the generic fallback ran the other way, so a third-party method implementing only `solve` was reached *through* `solve` from inside every iteration of a solve and emitted its messages there — with no `maxlog` left to bound it. `solve` is now one derived definition, the six copies and the "Solve method missing" stub are gone, `solve(ls, 1)` works, and the generic `solve_with_status` raises with the signature it owes instead of recursing. GeometricOptimizers' `DecayingStatic` implements both and is unaffected; its own `solve` stays more specific than the derived one, verified against a mock of that shape.
- **`Bisection` reports the merit it measured.** Two returns filled `φ` with `φ(0)` — no bracket at all, and a bisection that landed on a non-positive step — so for `φ(α) = 1 − α` the status read `α = 1.0, φ(α) = 1.0` where the value is `0.0`, and `linesearch_exhausted_reason` described a merit that descends forever as not having changed. `check_anchor` keeps `φ(0)` deliberately, and now says so.

The rest of that commit is documentation the first one made untrue: `solve_with_status` claiming only `Backtracking` reports a real outcome, the function-barrier rationale citing the `solver_step!` call the first commit deleted, three `maxlog` references in tests to caps that no longer exist, and a `Bisection` retry comment that reads as replacing the first attempt's verdict when the code keeps the worse of the two — left as it stands, since the two disagree only when φ′ is inconsistent with φ and `LINESEARCH_EXHAUSTED` says exactly that.

## Also in this PR

`CHANGELOG.md` gains an **Open Issues** section recording what surfaced during this work and was *not* fixed: that `Bisection` can still converge onto a *maximum* (the other route into an overclaimed `LINESEARCH_FLOOR`, which needs a change to what the method searches for rather than what it reports), that the trace options have zero readers, and that `should_report!`'s backoff does not promise every occurrence.

Raised in review and deliberately not changed: `flag_stall!` no longer fires for a failed bracket, so such a solve runs to `max_iterations` instead of stopping after `max_stalls` and relies on the step-based `stalled_step` — that is the point of the change, and the downstream check covers it. And `"Solver took N iterations."` still carries no `linesearch_reason`, because appending it re-raises the noise question above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,111 @@
 
 All notable changes to SimpleSolvers.jl are documented here.
 
+## [0.12.0]
+
+Two defects reported from downstream by
+[GeometricOptimizers.jl](https://github.com/JuliaGNI/GeometricOptimizers.jl), filed there as **D3**
+and **D4**. **Breaking**: a `NonlinearSolver` no longer emits line-search warnings from inside its
+iteration, and `NonlinearSolverStatus` gains a field.
+
+### `Bisection` no longer reports success when it cannot bracket
+
+#### The gap
+
+`_bisection_core` returned `(α, converged::Bool, n)` and set `converged = true` on the branch that
+handles *no sign change between the endpoints*:
+
+```julia
+if y₀ * y₁ > zero(y₀)                                  # no sign change
+    report_bisection_nobracket(α₀, α₁, y₀, y₁, config)
+    return (abs(y₀) ≤ abs(y₁) ? α₀ : α₁), true, n      # ← converged = true
+end
+```
+
+The intent was that a flat derivative at a minimum is benign. The effect was that the core could
+not tell *"found the root"* from *"gave up"*, and the endpoint of a bracket that never existed was
+handed on as if it were the located line minimiser.
+
+That claim then propagated. `Bisection` bisects ``\varphi'`` but brackets on ``\varphi``, so the
+two disagree exactly when something is wrong — a stale or regularized Jacobian, an inexact linear
+solve, a non-smooth merit. In that case the returned endpoint typically does not improve the merit
+either, and the search classified it as `LINESEARCH_FLOOR`: *"no line search can make progress
+here"*. The outer iteration acts on that (`flag_stall!`, then `max_stalls`), so a failure to
+bracket was being read as a property of the problem. Downstream it contributed to a diverging
+`_BFGS` + `Bisection` + `Geodesic` run in which the step was taken regardless.
+
+#### Changed
+
+- `_bisection_core` now returns a `BisectionOutcome` — `BISECTION_CONVERGED`,
+  `BISECTION_NOBRACKET` or `BISECTION_EXHAUSTED` — in place of the `Bool`. Internal, like the
+  `Symbol` that `_triple_point_core` returns. The `α` it returns is unchanged in every case.
+- `bisection` dispatches on it, so a spent budget and a failed bracket get the wording that fits
+  each (*"I did not narrow the interval far enough"* versus *"there is no root in it at all"*;
+  only the first is fixed by a larger budget).
+- The `Bisection` **line search** classifies a failed bracket by the merit alone, with
+  `LINESEARCH_FLOOR` disallowed: `LINESEARCH_DECREASED` when the returned step still beats
+  ``\varphi(0)`` by more than ``\tau``, `LINESEARCH_EXHAUSTED` when it does not. `LINESEARCH_FLOOR`
+  is now reachable only from a bisection that actually converged. The retry from the ``\alpha = 0``
+  anchor carries its own verdict through for the same reason; it used to be discarded.
+- The standalone `bisection`'s no-bracket warning moved from `verbosity ≥ 2` to `≥ 1`. It sat at 2
+  only because of the benign line-search case, and the line search no longer routes through
+  `bisection` — it calls `_bisection_core` and reports through its `LinesearchStatus`. For a
+  caller asking for a root, "there is no root in your interval" is a genuine failure.
+
+### A line search reports to its caller, not to the user
+
+#### The gap
+
+The `maxlog` caps on the line-search warnings were keyed on the *source location* of the `@warn`,
+so they were process-global and were never reset between `solve!` calls. Once spent, a message was
+gone for the rest of the session, including for later solves of entirely different problems — a
+genuine line-search failure in the tenth solve of a run was silent. This mattered because the log
+was the only channel by which a rejected line search was visible at all.
+
+The caps were a symptom. The channel split already existed — `solve_with_status` returns a
+`LinesearchStatus` and emits nothing, `solve` adds `linesearch_warnings` — but `solver_step!`
+called `solve_with_status` and then `linesearch_warnings` *anyway*, once per outer iteration. A
+solve that cannot make progress asks the line search for an impossible decrease at every one of its
+iterations, so that call is what turned one diagnosis into thousands of identical messages, and the
+caps existed to hold it back.
+
+#### Changed
+
+- `solver_step!` no longer calls `linesearch_warnings`. A line search now reports to its *caller*
+  through `LinesearchStatus`, and to the *user* only when the user called `solve(ls, α)` directly.
+  One direct call yields one message, so there is nothing left to rate limit and the three
+  `maxlog` caps in `report_linesearch_status` are **removed** rather than reimplemented.
+- One deliberate loss: a solver at `verbosity ≥ 2` no longer gets `curvature_diagnostic` per
+  iteration, which cost a full Jacobian evaluation each time.
+
+#### Added
+
+- `NonlinearSolverState` tallies the `LinesearchOutcome` of every step (`record_linesearch!`),
+  reset per solve by `initialize!` exactly like `stalls`.
+- `NonlinearSolverStatus` carries a snapshot of that tally, read with `linesearch_outcomes`,
+  `linesearch_failures` and `dominant_linesearch_outcome`. This is the programmatic channel the
+  package lacked: a caller that wants to *act* on a rejected line search — restart an approximate
+  Hessian, fall back to steepest descent — reads the status instead of scraping the log. Not
+  exported, per the convention the other status predicates follow.
+- The stagnation and no-progress messages of `nonlinear_solver_warnings` name what the line search
+  reported and how often ("the line search reported `LINESEARCH_NO_DESCENT` on 194 of the 200
+  step(s)"), so a failed solve names the cause and not only the symptom. A solve that converged
+  despite line-search failures says the same at `verbosity ≥ 2`, as an `@info`.
+
+### The solver's own report backs off instead of going silent
+
+`nonlinear_solver_warnings` fires once per `solve!` and is the one place a cross-solve cap is still
+needed — a time-stepping loop over an unattainable tolerance would otherwise report at every step.
+Its three repeatable messages are now gated on `should_report!`, which reports the 1st, 2nd, 4th,
+8th … occurrence of a diagnosis rather than the first three and then nothing ever again. Keys
+include the dominant line-search outcome, so a solve that starts failing for a *new* reason is
+reported at once.
+
+Be clear about the trade: a *repeating* diagnosis is reported on occurrences 1, 2, 4, 8, 16 …, so
+occurrence 10 is silent and occurrence 16 is not. Every occurrence is not promised — that would be
+the flood back. `verbosity = 0` still silences a solver completely, and the status is still the way
+to act on an outcome rather than read about it. `reset_warning_counts!` clears the counters.
+
 ## [0.11.0]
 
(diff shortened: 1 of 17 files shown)


michakraus (commented):
## Review

The design is right and I would merge it after the points below. The cut it makes — a line search
reports to its *caller* through `LinesearchStatus` and to the *user* only through a direct
`solve` — is the correct one, and it is what turns the `maxlog` problem from something to
reimplement into something to delete. `_bisection_core` returning a `BisectionOutcome` is the
minimal change that fixes D3, and keeping the returned `α` identical in every case makes it easy to
believe. The tally on `NonlinearSolverStatus` is the programmatic channel the package was missing,
and the `Open Issues` section is the right thing to have written down.

Verified locally against this branch, before any of the changes below:

- `Pkg.test()` green (1398 line-search + 3006 nonlinear-solver assertions).
- `docs/make.jl` clean, doctests pass, only the two pre-existing warnings.

Three of the findings are reproduced by running the package; the rest are documentation the change
made untrue. Fixes for all of them are prepared on the branch and are under review before they are
pushed.

---

### 1. The converged-solve `@info` fires on *every* healthy solve

`src/nonlinear/nonlinear_solver_status.jl:686` gates on `linesearch_failures(status) > 0`, and
`isbenign(LINESEARCH_FLOOR)` is `false` — so the terminal floor of a textbook solve trips it:

```julia
julia> x = ones(2); s = NewtonSolver(x, (y,x,p) -> y .= x.^2 .- 2, zero(x); verbosity=2);
julia> solve!(x, s)
[ Info: Nonlinear solver converged after 6 iterations to rfₐ = 6.28e-16, but not every step went
smoothly. The line search reported LINESEARCH_FLOOR on 1 of the 6 step(s), i.e. no trial step
changed the merit by more than its round-off resolution — the same floor this message reports one
level up.
```

Nothing about that solve was unsmooth, and the trailing clause ("the same floor this message
reports one level up") does not fit a *convergence* message. It also contradicts two docstrings
this PR writes: `linesearch_reason` says "the line search failed once" is noise from the last step
of an otherwise healthy solve, and `isbenign` promises the tally is named "only for a solve that
did *not* converge".

**Fix.** Count only the failures that are not the expected terminal floor. That needs the clause to
name the outcome that made the message fire, not the floor it is ignoring, so
`dominant_linesearch_outcome` gains a `count_floor` argument and `linesearch_reason` takes the
outcome to explain — otherwise a solve with one floor and one exhausted step fires because of the
latter and reports the former (the tie rule sends it to `LINESEARCH_FLOOR`).

### 2. `solve_with_status`'s own docstring is untrue about who reports

`src/linesearch/linesearch_status.jl:227`: *"Only `Backtracking` reports a genuine
`LinesearchOutcome`; the generic fallback calls `solve` …"*. All six built-ins implement
`solve_with_status`. This is the docstring a reader consults for the layering this PR establishes,
and it also hides the hole `Open Issues` records. Rewritten — see 7.

### 3. The barrier rationale in `report_linesearch_status` cites the call the PR deleted

Its docstring opens with "`linesearch_warnings` is called from `solver_step!` on every iteration of
every solve". The barrier is still needed — `solve` is specialized on the `Linesearch`, hence per
problem — so the rationale is restated rather than dropped. `test/linesearch_tests.jl:1247` repeats
the same claim.

### 4. Stale `maxlog` references left in test comments

`test/linesearch_tests.jl:1324,1327` ("a message the verbosity gate or `maxlog` discards";
"`TestLogger` records every message regardless of `maxlog`") and
`test/nonlinear_solver_tests.jl:1777` ("unlike the line-search messages"). None of the line-search
messages carries a cap any more.

### 5. Two small doc inaccuracies introduced here

- `src/SimpleSolvers.jl:93` lists the unexported accessors and omits `dominant_linesearch_outcome`.
- `nonlinear_solver_status.jl:632`: "Their keys include the dominant line-search outcome" is true of
  the two diagnoses, not of `:iterations`.

### 6. The `Bisection` retry comment describes something the code does not do

```julia
bretry === BISECTION_CONVERGED || (bres = bretry)   # bisection.jl:278-280
```

reads as "the retry's verdict replaces the first's", but it keeps the *first* attempt's verdict when
the retry converged — although `αres` then comes wholly from the retry. Reachable:

```julia
prob = LinesearchProblem{Float64}((a,_) -> (a + 2.0)^2, (a,_) -> a < -4.0 ? 1.0 : -1.0)
# first bracket (-3.0, 0.0)          -> BISECTION_NOBRACKET
# retry bracket (-5.11, -1.27)       -> BISECTION_CONVERGED
# status: LINESEARCH_EXHAUSTED       (the first attempt's verdict governs)
```

I left the behaviour alone and fixed the comment: that path needs `φ′` to be inconsistent with `φ`
(a derivative that agrees with the merit changes sign inside a bracket `bracket_minimum` found, so
the first bisection would have converged), and `LINESEARCH_EXHAUSTED` — whose reported reason is
exactly "φ′(0) is inconsistent with the merit" — is then the better of the two diagnoses. Worth
saying out loud, since keeping the worse verdict is not what the comment suggests.

---

## Two of the Open Issues closed as well

### 7. A `LinesearchMethod` now implements `solve_with_status`, and `solve` is derived (breaking)

The layering held for the six built-ins only because each one wrote the same three-line `solve` on
top of its `solve_with_status`. The generic fallback ran the other way, so a third-party method
implementing only `solve` was called *through* `solve` from inside every iteration of a solve and
emitted its messages there — the one thing the `LinesearchMethod` contract promises does not
happen, and now with no `maxlog` to bound it.

- `solve(ls::Linesearch, α, params)` becomes one derived definition; the six per-method copies and
  the "Solve method missing" stub go. `α` is converted to the `Linesearch`'s element type, so
  `solve(ls, 1)` works.
- The generic `solve_with_status` raises, naming the method and the signature it owes, instead of
  deriving a status from `solve` (the two would otherwise recurse).

This is breaking for a method that implements only `solve`. Checked downstream:
GeometricOptimizers' `DecayingStatic` (`src/optimizers/decaying_static.jl:76,84`) implements
*both*, and its own `solve` stays more specific than the new generic one — verified against a
mock of that exact shape.

### 8. `Bisection` reports the merit it measured

For `φ(α) = 1 − α` the status read `α = 1.0, φ(α) = 1.0` where the value is `0.0`, and
`linesearch_exhausted_reason` interpolates that difference — "the merit did not change" about a
merit nothing had evaluated. Two returns filled `φ` with `φ₀` (no bracket at all, and a bisection
that landed on a non-positive step; since this PR the latter can carry `LINESEARCH_EXHAUSTED`).
Both now evaluate at the step they hand back, at the cost of one merit evaluation on a path that
has already failed. The `check_anchor` returns keep `φ₀` deliberately, and now say so.

The third open issue (`Bisection` converging onto a *maximum*) is left as recorded: it changes what
the method searches for, not what it reports.

---

## Raised, not changed

- **`flag_stall!` no longer fires for a failed bracket.** That is the point of the PR, but it is
  worth stating as the behavioural risk: such a solve now runs to `max_iterations` instead of
  stopping after `max_stalls`, and relies on the step-based `stalled_step` to catch it. The
  downstream check covers it; a reader of the diff should not have to infer it.
- **"Solver took N iterations." carries no `linesearch_reason`.** A solve that spends its budget
  while still making slow progress therefore names no cause. Appending the clause would re-raise
  the noise question finding 1 answers (a healthy long solve floors on its last step), so it is a
  follow-up rather than a one-line addition.

After the changes: `Pkg.test()` green (1409 + 3017 assertions, Aqua included), docs and doctests
clean.

PR #181, opened 2026-08-14
Bound the step a line search may return

michakraus:
Fixes [GeometricOptimizers issue **D6**](https://github.com/JuliaGNI/GeometricOptimizers.jl/blob/main/CHANGELOG.md), the upstream half of their **A1b**, and documents every convergence and stopping criterion in one place.

## The defect

`Quadratic` and `BierlaireQuadratic` fit a polynomial to `φ` and step to its stationary point. The fits already confine that point to the bracket — what nothing confined was the **bracket**. `bracket_minimum_with_fixed_point` and `_triple_point_core` double their probe step up to `DEFAULT_BRACKETING_nmax = 100` times, so from `s = 1e-2` the right endpoint can reach `1e28`. Measured downstream on the `St(20,3)²` SVD problem, `Quadratic` returned `α = 4.3e7` on a direction of norm 5.54 — a step of `‖αδ‖ = 2.4e8` — and did it again two steps later on a steepest-descent direction. `bracket_minimum`, and hence `Bisection`, has the same shape.

Why it went unreported is the part that makes it ours rather than downstream's. In a Euclidean problem it is self-correcting: `f(x + αp)` grows with `α`, so the search's own decrease test throws the step out. On a **compact** manifold it is not — `φ` is bounded there and at `α = 1e9` can be genuinely *lower* than at `α = 0`, so the search reports `LINESEARCH_DECREASED` on a step nine orders of magnitude too long and is telling the truth. **The merit is not a bound on the step.**

## The fix

Every method now returns `α ≤ αmax`, a sixth clause of the contract in `LinesearchMethod`, with two ways to set the ceiling:

- **The method's own** — the `αmax` field, generalised from `StrongWolfe` (which has carried exactly this field, with exactly this meaning, all along) to `Bisection`, `Quadratic` and `BierlaireQuadratic`. All four default to `DEFAULT_LINESEARCH_αmax`, which *is* `DEFAULT_WOLFE_αmax = 65536`; the latter is now defined as the former so they cannot drift.
- **The caller's** — an optional `αmax` field of `params`:
  ```julia
  solve_with_status(ls, one(T), (x = x, state = state, αmax = 2π * c / norm(δ)))
  ```
  read through the same compile-time `hasproperty` guard as `params.φ₀`. Not a keyword, so `solve_with_status(ls, α, params)` keeps the signature the contract names as the extension point — GeometricOptimizers' `DecayingStatic` is unaffected.

The caller's half has to be **per call** (the scale changes with `‖δ‖` every step) and has to be an **input** rather than a clamp on the result: clamping afterwards yields a step the search never evaluated, whose `LinesearchStatus` then describes a different step than the one taken. As an input the bracketing stops at the ceiling — so the evaluations beyond it are saved, not wasted — the merit is measured there, and the status describes what is handed back. The bound applies to *where the merit is evaluated* too, including each bracketer's first probe.

All six methods honour it, including the two with no ceiling of their own: `Backtracking` clamps its trial step and bounds its expansion phase, `Static` clamps its fixed `α`, and `check_anchor` clamps so the ceiling holds on the returns that never search.

**No new `LinesearchOutcome`.** A capped search that decreased the merit *did* decrease it; `capped_status` classifies it by the same `τ` rule as any other step. A `LINESEARCH_CAPPED` would have changed `NLINESEARCH_OUTCOMES`, the tally in `NonlinearSolverState` and every downstream read of it, to say something a caller who supplied a ceiling already knows.

## Behaviour

One case changes at default options: a merit that keeps decreasing to the right used to exhaust the bracketing budget and report `LINESEARCH_EXHAUSTED` with the caller's `α`; it now returns the ceiling as `LINESEARCH_DECREASED`, because the merit there really is lower. One test expectation moved with it, and the old behaviour is asserted alongside the new one at `αmax = Inf`.

Nothing else moves unless the ceiling binds. The six per-method evaluation canaries and every zero-allocation assertion stand unedited, a ceiling in `params` allocates nothing, and `Bisection` gains a field while keeping all four of its constructor forms.

## Measured

GeometricOptimizers' SVD problem, `_BFGS` + `Cayley`, over the eight starting points of `scripts/retraction_accuracy.jl` (cap 20 000). *On the manifold* means `check ≤ 1e-12`, that package's own `MANIFOLD_TOLERANCE`:

| search | ceiling | seeds ending on the manifold | worst `check` |
|---|---|---|---|
| `Quadratic` | none (before) | 4 of 8 | 1.3 |
| `Quadratic` | 65536 (the default) | 4 of 8 | 2.8 |
| `Quadratic` | 10 | 7 of 8 | 4.1e-9 |
| `Quadratic` | 1 | **8 of 8** | **6.2e-14** |
| `BierlaireQuadratic` | none (before) | 2 of 8 | 0.92 |
| `BierlaireQuadratic` | 65536 (the default) | 4 of 8 | 0.38 |
| `BierlaireQuadratic` | 10 | 7 of 8 | 4.7e-9 |
| `BierlaireQuadratic` | 1 | **8 of 8** | **6.4e-14** |

**The default ceiling does not close A1b, and the CHANGELOG says so.** At `‖δ‖ ≈ 5.5` a step of `α = 65536` is still `‖αδ‖ = 3.6e5`, five orders above the `2π` past which retracting a lift only adds round-off. What the default reliably does is remove the unbounded extrapolation and bound the bracketing cost. What *fixes* A1b is a ceiling of the right **magnitude** — the bottom row of each block — which is precisely why the caller's half is not optional. That ceiling is GeometricOptimizers' to choose; this PR is what lets them pass one.

Everything else downstream is unmoved: GeometricIntegrators' RK suite passes (107 assertions) and `LotkaVolterra2d` over 1000 steps is **bit-identical** for `Gauss(1)`, `Gauss(2)`, `Gauss(3)`, `ImplicitMidpoint` and `SRK3`. On the SVD sweep exactly three rows change, all under `Cayley`; every `Geodesic` and Euclidean-scaled row is identical to the digit.

## Also in this PR

**`docs/src/convergence.md`**, a new page. The criteria were spread across a dozen docstrings on unexported functions, so there was no statement of them a reader could follow end to end. It covers the line search (`τ`, all six outcomes with the test that produces each, the anchor policy, `αmin`, `αmax`, the six contract clauses), the nonlinear solver (the three residuals, the shared `residual_small` gate that both convergence branches require and both give-up branches negate — which is *why* converging and stagnating are mutually exclusive — `assess_convergence`, the four ways a solve ends without converging, the full `meets_stopping_criteria` disjunction), and **the channel between them**: how a `LinesearchStatus` becomes a `flag_stall!`, a forced Jacobian refresh, a per-solve tally, and the clause that names the cause in the solver's own failure message.

**CHANGELOG `## Open Issues`** gains what this work turned up and did not fix — chiefly that A1b stays open until GeometricOptimizers passes its own ceiling, that a binding ceiling leaves no trace in the status, and that the capped path costs one merit evaluation the bracketing already made.

## Testing

- Full suite passes: **1562** line-search and **3017** nonlinear-solver assertions, plus Aqua and JET.
- New coverage: the reproducer at the default ceiling, the caller channel across all six methods, a ceiling binding below the natural minimiser, the no-change case (step, outcome *and* evaluation count identical with a generous ceiling), input validation before any evaluation, the evaluation bound, the `:capped` bracketing units, and the allocation canary with a ceiling in `params`.
- Docs build with zero cross-reference errors.

---

## Update: `Bisection` converges onto a minimum, never onto a maximum

Also fixes the **`Bisection` can converge onto a maximum** entry that stood under `## Open Issues` — the second of the two routes by which `Bisection` could claim `LINESEARCH_FLOOR` without having earned it. 0.12.0 closed the route through a *failed* bracket; this closes the route through a **successful** one.

### The gap

`Bisection` drives on the sign of `φ′`, so it converges to whichever crossing the sign of `φ′` at its left endpoint selects — and **that sign is invariant under the halving**: `y₀` is reassigned only on the branch where `y₀ * y > 0`, so it never changes sign, and `y₁` keeps the opposite one throughout. From `φ′(lo) < 0` the interval shrinks onto a `- → +` crossing, a minimum; from `φ′(lo) > 0` onto a `+ → -` crossing, a **maximum**.

Nothing ruled the second out. `bracket_minimum` brackets a minimum in **value** — it samples `φ` and never `φ′` — so on a non-convex ray its interval can enclose several stationary points and its left endpoint can sit past one of them. Landing on a maximum was not caught: the step was classified by the merit like any other, so it came back as `LINESEARCH_DECREASED` when it happened to improve `φ` and as `LINESEARCH_FLOOR` when it did not. The second is the overclaim, because `LINESEARCH_FLOOR` is a claim about the *direction* — that no line search can progress along it — which the outer iteration acts on through `flag_stall!` and `max_stalls`. GeometricOptimizers observed exactly this (`_BFGS` + `Bisection` + `Geodesic`: `LINESEARCH_FLOOR` with `φ(1) = φ(0)`, and the step taken regardless).

### The fix

- `_bisection_core` returns the value at its (sorted) **left endpoint** as a fourth element — returned rather than recomputed, because its sign is the invariant above and so names which crossing the bisection is heading for before it has finished heading there.
- `_bisect_for_minimum` reads it. Where `φ′(lo) > 0` it discards the interval and bisects `[0, lo]`, which brackets a minimum **by construction** — `check_anchor` has established `φ′(0) < 0`, and `φ′(lo) > 0` supplies the opposite sign at the other end. The minimum it finds is the *earlier* one, nearer the anchor, which is the one a line search wants. The same condition repairs a `BISECTION_NOBRACKET` whose interval ascends at `lo`.
- The repair's verdict **replaces** the first bisection's rather than being merged with it, unlike the negative-step retry: the two disagree because the bracket was in the wrong *place*, which is what the orientation detected — not because `φ′` is inconsistent with `φ`.

### Cost

**None on a ray that does not need it.** The orientation is read off a value `_bisection_core` computes anyway, so a correctly oriented bracket is recognised without one extra evaluation of `φ′` — a full Jacobian for the `‖F‖²` merit of a `NonlinearSolver`. Asserted directly: over three intervals the repaired path returns the same step, the same verdict and the same evaluation count, and makes the same number of derivative evaluations. The overshoot branch was always oriented by construction and now says so through the same helper rather than by implication.

### Measured

The downstream symptom, reproduced: on `φ(α) = (α-1)²` with `φ′` given its own shape (minimum at `0.3`, maximum at `2.0`), `bracket_minimum` returns `[0.64, 2.56]`, which *ascends* at `0.64` and *descends* at `2.56`. Bisecting it lands on `α = 2.0` — the maximum — where `φ(2) = φ(0)` exactly, i.e. `LINESEARCH_FLOOR`. Repaired, the search returns `α = 0.3` with `LINESEARCH_DECREASED` and `φ = 0.49` against `φ(0) = 1.0`.

And the orientation is not exotic. Over 300 random starting points on the non-convex `exp(x)(x³ − 5x² + 2x) + 2` root-finding fixture with `Bisection` as the `NewtonSolver` line search, **13 of 300 solves now converge to a different root**. Not a worse one — all 300 converge in both cases, the same eight distinct roots are reached, and the distribution of `|F|` at the returned iterate is **bit-identical** (median `6.7e-16`, p95 `1.4e-13`, worst `2.7e-12`, all below `1e-10`). Total iterations drop from 698 to 687, worst case unchanged at 4. On a merit with one minimum nothing moves at all.

Full suite still passes; docs still build with zero cross-reference errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,34 @@ divided by `‖δ‖` — is not something the search can measure.
   `bracket_minimum`, which reports a truncated bracket as an ordinary one — it was the one path in
   the method that could not reach `capped_status`, and would have bisected `φ′` over an interval on
   which it keeps its sign.
+- A ceiling at or *below* the bracketing start is answered without a search. `_triple_point_core`
+  bounds its first probe with `δ = min(δ, αmax - x₀)`, which is not positive when `αmax ≤ x₀`, so
+  every probe landed at or *left* of the start: on `f(α) = 1 - α`, `αmax == x₀` reported `:flat`
+  after two evaluations — that no line search can decrease the merit here, about a search that was
+  never allowed to look — and `αmax < x₀` spent twelve halving a negative `δ` first. It now
+  reports `:capped` for no evaluations at all. `BierlaireQuadratic` guarded this before it called,
+  which is why nothing leaked; the guard belongs where the `αmax` argument is.
+- The two optional `params` fields are read through the same guard. `φ₀` used `haskey` while
+  `αmax` used `hasproperty`, which agree on a `NamedTuple` and on nothing else: a plain struct had
+  its ceiling honoured and then raised a `MethodError` from inside the merit. Both use
+  `hasproperty` now, which is what the rest of the closure requires anyway — it reaches `params.x`
+  and `params.parameters` by property access, so a `Dict`, the one thing `haskey` bought, could
+  never have worked. `NullParameters` and any struct do.
+- `Quadratic` no longer pays a derivative evaluation for a search the ceiling then cancels. It
+  chose its bracketing start — the `φ′(α₀)` of issue #164 — before testing whether the ceiling
+  left anything to search, so a trial step at or beyond the ceiling, which is the common case
+  once a caller supplies one, cost a full `Jacobian` for the ``\|F\|^2`` merit and discarded the
+  answer. The ceiling is tested first; the answer is the same on either side of the minimiser.
+- `trials` counts what the search really spent. The bracketing cores now report their evaluation
+  count alongside their status, and the three searches that bracket carry it into the
+  [`LinesearchStatus`](@ref) — as does `capped_status`, and as do the four returns that measure
+  the merit at the step they hand back. The field was documented as a lower bound for those
+  methods, and on the capped path, where the bracketing *is* the search, the bound was vacuous:
+  `Bisection` reported `1` after eight merit evaluations and `Quadratic` reported `0` after two,
+  which is the value `Static` reports for evaluating nothing at all. `Quadratic`'s fixed `n += 2`
+  per fit round goes with it — it was a proxy for the two endpoint merits from when the
+  bracketer's cost was invisible, and those are exactly the evaluations now counted for real.
+  `Backtracking` and `StrongWolfe` are unchanged and still exact.
 
 #### Changed (behavioural)
 
@@ -1382,54 +1410,6 @@ follows is what is left.
   visible where it previously took a contrived merit to reach. Whether a Wolfe search *should* call
   a step that was never allowed to grow "exhausted" is a question about that method, not about the
   ceiling, so it was left alone.
-
-### Raised while reviewing the step ceiling, not addressed
-
-The review that found the truncation defect (`_bracket_core` reporting a bracket clamped to the
-ceiling as a found one, fixed in 0.12.0) turned up four more, none of which it closed. The
-`LINESEARCH_FLOOR` entry above is a fifth and belongs to this list as much as to the one it sits in.
-
-- **`_triple_point_core` walks the wrong way for a ceiling at or below its start.** It bounds its
-  first probe with `δ = min(δ, αmax - x₀)`, which is not positive when `αmax ≤ x₀`, so `x₁ = x₀ + δ`
-  lands at or *left* of the start and the rest of the routine is nonsense: measured on
-  `f(α) = 1 - α`, `αmax == x₀` reports `:flat` after two evaluations — which
-  `BierlaireQuadratic` would map to `LINESEARCH_FLOOR`, a claim that no line search can progress
-  along the direction — and `αmax < x₀` spends twelve evaluations halving a negative `δ` before
-  reporting `:unbracketable`. Unreachable from a line search, because `BierlaireQuadratic` returns
-  `capped_status` on `start < αmax` failing before it calls the core, and that guard is the reason
-  the core was left without one of its own. Reachable by any other caller of `_triple_point_core`,
-  which is where the `αmax` argument lives. The fix is a `δ > 0` check in the core rather than in
-  its one caller; left because adding it means deciding what a core asked to search a range of zero
-  width should say, and `:capped` — the honest answer — is a status the public
-  `triple_point_finder` cannot express.
-- **The two `params` channels are not read the same way, and only one of them accepts a struct.**
-  `caller_αmax` guards with `hasproperty`, while the merit closure of `linesearch_problem` guards
-  `φ₀` with `haskey`. They agree on a `NamedTuple`, which is what every caller in this package and
-  downstream passes, so nothing is broken today. They do not agree on anything else: for a plain
-  struct with `x`, `αmax` and `φ₀` fields, `hasproperty` finds the ceiling and honours it while
-  `haskey` raises a `MethodError` from inside the merit — verified. So the two halves of what the
-  docstrings describe as one channel impose different requirements on `params`, and the stricter of
-  them is undocumented. `hasproperty` throughout would be the wider contract and is a one-line
-  change; left because widening what `params` may be is a decision about the extension point, not
-  about the ceiling, and it wants its own tests for the `NullParameters` and `Dict` cases.
-- **`Quadratic` pays a derivative evaluation for a search the ceiling then cancels.**
-  `_quadratic_search` picks its bracketing start with `derivative(problem(ls), α₀, params)` and only
-  *then* tests `α < αmax`, so a trial step at or above the ceiling — the common case once a caller
-  supplies one, since the ceiling is what makes the trial step inadmissible — costs a derivative
-  evaluation whose answer is discarded. Measured on `φ(α) = 1 - α` with `params.αmax = 0.5`: two
-  derivative evaluations for a search that returns the ceiling without bracketing. For the
-  ``\|F\|^2`` merit of a `NonlinearSolver` that is a full `Jacobian`. Hoisting the ceiling test
-  above the start selection fixes it; left because the ordering interacts with issue #164's rule
-  for *where* the bracketing starts, and untangling the two is more than the one evaluation is
-  worth on its own.
-- **`trials` says almost nothing about a capped search.** The field is documented as a lower bound
-  for the bracketing searches, since evaluations spent inside a bracketing helper are not counted —
-  but on the capped path the bracketing *is* the whole search, so the bound is vacuous. Measured:
-  `Bisection` reports `trials = 1` after eight merit evaluations, and `Quadratic` reports
-  `trials = 0` after two, which is the same value `Static` reports for a method that evaluates
-  nothing at all. Nothing reads `trials` except the warning messages, which is why this is an
-  entry and not a defect, but a caller sizing the cost of a line search from it is misled exactly
-  where the ceiling made the search cheap.
 
 ### Raised while fixing the `Bisection` maximum, not addressed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1337,6 +1337,46 @@ follows is what is left.
   a step that was never allowed to grow "exhausted" is a question about that method, not about the
   ceiling, so it was left alone.
 
+### Raised while fixing the `Bisection` maximum, not addressed
+
+- **The orientation is not checked on a bracket that `bracket_minimum` flipped.**
+  `_bisect_for_minimum` returns early on `lo ≤ 0`, which covers two different things: the overshoot
+  branch, where `lo = 0` and `ylo` *is* the `φ′(0) < 0` the anchor check established (genuinely
+  nothing to check), and a bracket that the walk flipped leftward, where `lo < 0` and nothing has
+  been established at all. The second is a real hole, and it is demonstrable: on `φ(α) = α²` with
+  `φ′` given the roots `-0.5`, `0.3`, `0.7`, `bracket_minimum` returns `[-1, 1]` with `φ′(-1) > 0`,
+  and the bisection converges to `α = -0.5` — a **maximum**, reached with the check skipped.
+
+  Nothing leaked in that instance, because `-0.5 < 0` and the α > 0 contract rejected it: the
+  negative-step retry fired and the search reported `LINESEARCH_EXHAUSTED`. Whether the same route
+  can select a maximum at *positive* α — the interval spans zero and may contain several `+ → -`
+  crossings, only some of them negative — was **not** established either way. It is hard to arrange
+  because `check_anchor` guarantees `φ′(0) < 0`, so a crossing selected from a positive left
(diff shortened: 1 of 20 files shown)


michakraus (commented):
## Review

I read the whole diff and exercised it against the built package. The design is right: two halves,
the caller's one riding on `params` rather than on a new keyword so that the `solve_with_status(ls,
α, params)` extension point survives; `capped_status` as a shared definition rather than a new
`LinesearchOutcome`; bounding *where the merit is evaluated* and not only what is returned. The
`Bisection` orientation fix and `docs/src/convergence.md` are both good, and the CHANGELOG's Open
Issues section is unusually honest about what the default ceiling does not close.

The full suite is green on the branch as submitted (1584 line-search + 3017 nonlinear-solver
assertions). But there is one defect that defeats the caller's half — the half the PR argues is the
one that actually closes A1b — for `Bisection`, at exactly the ceilings the measurements identify as
the ones that matter. Everything below is fixed in the follow-up commit on this branch, except where
I say otherwise.

---

### 1. A ceiling at or below the bracketing step is stepped past, not reported (major)

Every bracketer clamps its **first** probe to `αmax` — that is the fix in
`_bracket_core`/`_bracket_minimum_core`/`_bracket_minimum_with_fixed_point_core`, and it is right.
But when the ceiling is the smaller of the two, that clamped probe *is* the ceiling, so the loop's
probe lands on the same point:

```julia
b = s > zero(T) ? min(a + s, αmax) : a + s   # b == αmax when s ≥ αmax - a
...
if s > zero(T) && c ≥ αmax
    c = αmax                                  # ... c == b
    c > a || return (a, a, :capped)
    yc = f(c)                                 # the same point, evaluated again
    return (a, c, bc(yb, yc) ? :ok : :capped) # yc == yb, and `yc ≥ yb` is TRUE
end
```

`BracketMinimumCriterion` is `yc ≥ yb`, so a tie satisfies it. The truncation is reported as **`:ok`**
— a turning point that is one point counted twice — and the whole `:capped` route, `capped_status`
included, becomes unreachable below `s`.

For `Bisection` that is not a corner case: it seeds `s = clamp(|α|, 0.01, 1)`, which is `1` for the
usual trial step `α = 1`. So it fires for **every caller ceiling ≤ 1** — and the PR's own table puts
`αmax = 1` as the row that takes the SVD sweep from 4-of-8 to 8-of-8. Measured on
`φ(α) = 1 − α` (falls monotonically, so the ceiling is by construction the best admissible step):

| `params.αmax` | α returned | best admissible | outcome | merit evals |
|---|---|---|---|---|
| 2.0   | 2.0      | 2.0   | `DECREASED` | 6 |
| 1.0   | **0.32** | 1.0   | `DECREASED` | 16 |
| 0.5   | **0.16** | 0.5   | `DECREASED` | 15 |
| 0.05  | **0.02** | 0.05  | `DECREASED` | 12 |
| 0.005 | 0.005    | 0.005 | **`EXHAUSTED`** | 10 |

Above the boundary it is correct and costs 6 evaluations. At or below it the search returns roughly a
third of the step it was allowed, spends 12–16 merit evaluations instead of ~4, evaluates the ceiling
4–7 times, and at the bottom row reports a **failure** — `LINESEARCH_EXHAUSTED` is counted by
`linesearch_failures` and named in the solver's own failure message — for a search that found a
decrease. The route it takes instead is `bracket → spurious :ok on [0, αmax] → φ′ has no sign change
→ BISECTION_NOBRACKET → the αres ≤ 0 retry → some interior point`.

`Quadratic` masks it: `_quadratic_search` short-circuits with `α < αmax || return (αmax, n)` whenever
the trial step is at or above the ceiling. `_bracket_minimum_with_fixed_point_core` has the identical
tie, and it fires when the trial step is *below* a ceiling ≤ `s`. `_triple_point_core` classifies it
correctly — it tests `xₖ₊₁ > xₖ` alongside the rise — but still pays for the duplicate evaluation.

**Fixed** by guarding the re-probe against the previous probe rather than only against the left
endpoint (`c > b`, `αmax > b`, `xₖ₊₁ > xₖ`), and by having `_bracket_minimum_core` return the
truncation directly instead of handing a ceiling it has already reached to `_bracket_core` — which
also removes the second of the two duplicate evaluations at the ceiling. All five rows now return the
ceiling as `LINESEARCH_DECREASED` for four evaluations. Nothing on the default path moves: with
`αmax = Inf` the branch is dead, and with the method's own `65536` the first probe is never the
ceiling.

### 2. `Bisection`'s negative-step retry cannot reach `capped_status` (minor)

The `αres ≤ 0` retry brackets through the **public** `bracket_minimum`, which by design reports a
truncated bracket as an ordinary one. It is the one path in the method that skips the
`bstatus === :capped` handling the first bracketing gets, and it would bisect `φ′` over an interval
on which `φ′` keeps its sign. **Fixed** — it goes through `_bracket_minimum_core` now, with `:capped`
treated as it is above.

### 3. The default ceiling silently disappears in `Float16` (minor)

`DEFAULT_LINESEARCH_αmax` is `65536`; `floatmax(Float16)` is `65504`. So `T(DEFAULT_LINESEARCH_αmax)`
is `Inf`, and `Bisection(Float16).αmax === Inf` — the defect the field exists to fix is absent in the
one precision `armijo_ulps`, `linesearch_iterations` and `default_precision` all go out of their way
to special-case, and which the suite constructs (`BierlaireQuadratic(Float16)`). Pre-existing for
`StrongWolfe`, newly inherited by three more methods and now a clause of the contract.

**Fixed** with `default_linesearch_αmax(T)`, saturating at `floatmax(T)`, used as the constructor
default in all four methods; `change_precision` saturates a finite ceiling through `convert_αmax` and
leaves an explicit `Inf` alone, since that one is a statement rather than an overflow.

### 4. A binding ceiling can be reported as `LINESEARCH_FLOOR` (not changed — recorded)

`capped_status` classifies the step at `αmax` by the same `τ` rule as any other, so a merit that is
still falling at the ceiling — which is what `:capped` *means* — but has fallen by less than `τ` over
the whole admissible range comes back as a floor. `LINESEARCH_FLOOR` is a claim about the
**direction**, and the outer iteration acts on it via `flag_stall!`/`max_stalls`; what was actually
established is only that no step the caller *permits* decreases the merit measurably. It is the same
shape as the two unearned floors this PR series removed from `Bisection`, reached through a third
door, and it is reachable precisely where the caller's ceiling is tightest.

I left it alone, because `LINESEARCH_EXHAUSTED` says "no step was found", which is false of a step
that was found and returned — closing it properly is the boolean-on-the-status already listed under
Open Issues. Added there with the reasoning.

### 5. Documentation (fixed)

- `quadratic_bierlaire.jl` refers to `_capped_status`; the function is `capped_status`.
- `linesearch_αmax`'s example passes `state = state`, which nothing reads — the closures in
  `linesearch_problem` read `params.x` and `params.parameters`. Same in
  `docs/src/linesearch/linesearch.md`, which omits `parameters` altogether.
- `linesearch_problem`: "Both are read through a `hasproperty` guard" — `φ₀` is read with `haskey`;
  only `αmax` uses `hasproperty`.
- Contract clause 1 ("It never throws") in `LinesearchMethod` and in `convergence.md` needs the
  cross-reference to the deliberate `ArgumentError`, which `linesearch_αmax` itself already calls
  "the one deliberate exception".
- `triple_point_finder`'s docstring documents `:capped`, but the public wrapper takes no `αmax` and
  can never return it.
- `bracket` / `bracket_minimum_with_fixed_point` signature lines and `bracket_minimum`'s
  `# Arguments` omit `αmax`.
- `Static` and `Backtracking` say nothing about `params.αmax` clamping them; only source comments do,
  and for `Static` the clamp is the *only* thing that can bind.

---

### Tests added

The ceiling-at-or-below-the-first-probe case for all three cores, with a probe count asserting the
ceiling is evaluated once; the five-ceiling end-to-end table for all three minimising searches on a
monotonically-falling merit; a turning point exactly *at* the ceiling still reporting `:ok`; and the
`Float16` ceiling being finite through both the constructors and `change_precision`.

Suite after the fixes: **1705** line-search (from 1584) and **3017** nonlinear-solver assertions,
plus Aqua. Docs build unchanged.

PR #182, opened 2026-08-15
Let `solve_with_status!` reuse the caller's state

michakraus:
Targets **0.12.1**. Additive: the one form of `solve_with_status!` that a solve in a loop actually wants.

## The gap

0.12.0 made the status the channel by which a rejected line search reaches its caller — `solve!` no longer warns from inside its iteration, so a caller that wants to know reads `NonlinearSolverStatus`, and `solve_with_status!` is how it does that without holding a `NonlinearSolverState`.

But both of its methods *build* that state:

```julia
solve_with_status!(x, s, params)                # allocates a NonlinearSolverState
solve_with_status!(x, prob, method, params)     # allocates a solver and a state
```

while `solve!` has had a state-taking method for as long as the state has existed, and the 0.11.0 docstrings tell a caller in a loop to build one solver and reuse it. So the caller with the strongest reason to read a status — a time-stepping loop, which runs one solve per step and is exactly where a per-solve allocation is worth avoiding — was the one caller `solve_with_status!` had nothing for. It had to write the two lines out by hand:

```julia
solve!(x, s, state, params)
status(s, state)
```

which is not merely inconvenient. `status(s, state)` is unexported, so the pattern reaches past the public surface for something the package already computes; and the two lines can drift apart, which is the failure `solve_with_status!` exists to prevent.

Reported downstream by [GeometricIntegrators.jl](https://github.com/JuliaGNI/GeometricIntegrators.jl), where **22 of the 33 `NonlinearSolver` call sites** are of this shape, every one of them inside an `integrate_step!`.

## Added

- `solve_with_status!(x, s, state, params=NullParameters())`. It is the two lines above, and the existing `solve_with_status!(x, s, params)` is now one line on top of it rather than a second copy of them — the same collapse the 0.11.0 constructors made, and for the same reason.
- The state is left holding the outcome, so a later `status(s, state)` returns what the call returned. That is asserted rather than implied: the two readings cannot disagree, and a second solve through the same state reports the same outcome, which is what "reuse" has to mean.

## Risk

Nothing else moves. The added method is strictly more specific than the `params` one it sits beside (`NonlinearSolverState` against the untyped `params`), so no existing call changes which method it reaches, and no behaviour of any of the three forms changes.

## Verification

- Full test suite passes, including the new assertions, over all four solver methods (`Newton`, `QuasiNewton`, `Picard`, `DogLeg`) and both `Float64` and `Float32` — 3057 assertions in *Nonlinear Solvers* alone. Run again by the pre-push hook.
- `Documenter.doctest(SimpleSolvers)` passes; the new signature is documented with a runnable example of the reuse.
- Downstream, GeometricIntegratorsBase and GeometricIntegrators were migrated onto it and their suites pass. `LotkaVolterra2d` over 1000 steps is **bit-identical** before and after for `Gauss(1)`, `Gauss(2)`, `Gauss(3)`, `ImplicitMidpoint` and `SRK3`, compared as raw `Float64` bit patterns rather than printed digits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,10 +50,28 @@ shape, every one of them inside an `integrate_step!`.
 - The state is left holding the outcome, so a later `status(s, state)` returns what the call
   returned. That is asserted rather than implied: the two readings cannot disagree, and a second
   solve through the same state reports the same outcome, which is what "reuse" has to mean.
+- `solve_with_status!(x, prob, method, args...)` now forwards its positional arguments the way
+  `solve!(x, prob, method, args...)` always has, so the state form is reachable through the
+  problem-taking wrapper too. `solve!`'s docstring documents those arguments as "either nothing at
+  all, the `params` of the problem, or a `NonlinearSolverState` followed by `params`"; a caller who
+  carried that sentence across to `solve_with_status!` used to get a `MethodError`.
 
-Nothing else moves. The added method is strictly more specific than the `params` one it sits beside
-(`NonlinearSolverState` against the untyped `params`), so no existing call changes which method it
-reaches, and no behaviour of any of the three forms changes.
+### The status is built once per solve
+
+`solve!` builds a `NonlinearSolverStatus` at the end of its loop to report on itself, and
+`solve_with_status!` used to ask `status(s, state)` for a second one afterwards — five `l2norm`
+passes per solve spent reproducing a value that had just been discarded, on precisely the path
+whose reason for existing is to keep per-solve work out of a caller's loop.
+
+Both now share one body, `SimpleSolvers._solve_core!`, which returns the status the loop already
+built; `solve!` returns `x` and `solve_with_status!` returns the status. The value is the same one
+either way, and the assertion above — that `status(s, state)` afterwards agrees with what the call
+returned — is what keeps it so.
+
+Dispatch does not move, and no form's observable behaviour changes. The added method is strictly
+more specific than the `params` one it sits beside (`NonlinearSolverState` against the untyped
+`params`), so no existing call changes which method it reaches; the `args...` wrapper accepts
+strictly more than the `params` one it replaces; and `solve!` returns what it always did.
 
 ## [0.12.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,59 @@
 
 All notable changes to SimpleSolvers.jl are documented here.
 
+## [0.12.1]
+
+Additive: the one form of `solve_with_status!` that a solve in a loop actually wants.
+
+### `solve_with_status!` can reuse the caller's state
+
+Reported from downstream by
+[GeometricIntegrators.jl](https://github.com/JuliaGNI/GeometricIntegrators.jl).
+
+#### The gap
+
+0.12.0 made the status the channel by which a rejected line search reaches its caller — `solve!` no
+longer warns from inside its iteration, so a caller that wants to know reads
+`NonlinearSolverStatus`. `solve_with_status!` is how it does that without holding a
+`NonlinearSolverState`.
+
+But both of its methods *build* the state:
+
+```julia
+solve_with_status!(x, s, params)                # allocates a NonlinearSolverState
+solve_with_status!(x, prob, method, params)     # allocates a solver and a state
+```
+
+whereas `solve!` has had a state-taking method since the state existed, and the docstrings tell a
+caller in a loop to build one solver and reuse it. So the caller with the strongest reason to read
+a status — a time-stepping loop, which runs one solve per step and is exactly where a per-solve
+allocation is worth avoiding — was the one caller `solve_with_status!` had nothing for. It had to
+write the two lines out by hand:
+
+```julia
+solve!(x, s, state, params)
+status(s, state)
+```
+
+which is not merely inconvenient. `status(s, state)` is unexported, so the pattern reaches past the
+public surface for something the package already computes; and the two lines can drift apart, which
+is the failure `solve_with_status!` exists to prevent. Measured downstream: 22 of the 33
+`NonlinearSolver` call sites in GeometricIntegrators and GeometricIntegratorsBase are of this
+shape, every one of them inside an `integrate_step!`.
+
+#### Added
+
+- `solve_with_status!(x, s, state, params=NullParameters())`. It is the two lines above, and the
+  existing `solve_with_status!(x, s, params)` is now one line on top of it rather than a second
+  copy of them — the same collapse the 0.11.0 constructors made, and for the same reason.
+- The state is left holding the outcome, so a later `status(s, state)` returns what the call
+  returned. That is asserted rather than implied: the two readings cannot disagree, and a second
+  solve through the same state reports the same outcome, which is what "reuse" has to mean.
+
+Nothing else moves. The added method is strictly more specific than the `params` one it sits beside
+(`NonlinearSolverState` against the untyped `params`), so no existing call changes which method it
+reaches, and no behaviour of any of the three forms changes.
+
 ## [0.12.0]
 
 Three defects reported from downstream by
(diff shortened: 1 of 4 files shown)


michakraus (commented):
## Review

The change is correct as written. I checked the three things that could plausibly have gone wrong here, and none did:

**Dispatch.** At three arguments `(AbstractArray, NonlinearSolver, NonlinearSolverState)` is strictly more specific than `(AbstractArray, NonlinearSolver, Any)`; at four, the new method and the `prob`/`method` one are disjoint in positions 2–3. `Test.detect_ambiguities` is empty and no existing call changes which method it reaches, as the description claims.

**The equality assertions are real, not tautological.** This was my first suspicion — `@test status(sv, statev) == stv` looks like it could be comparing an object with itself. It is not: `NonlinearSolverStatus` is an immutable, all-isbits struct, so the `==` fallback is `===`, i.e. a field-by-field bit comparison, and `status(s, state)` constructs a *fresh* status. So the assertion compares two independently built values, and it compares them bitwise — which is the strongest form this test could take.

**The second-solve assertion is not flaky.** It asserts determinism across four solver methods, which is only safe if every piece of cross-solve state is reset. It is: `initialize!(state, …)` clears `stalls`/`stallflag`/the progress reference/the line-search tally (`nonlinear_solver_state.jl:100`), `initialize!(cache(s), x)` clears the cache and — for `DogLeg`, the one method carrying a genuinely path-dependent scalar — resets the trust radius to `DOGLEG_Δ_INITIAL` (`dogleg_cache.jl:72`), and `maybe_refactorize!` refreshes unconditionally at `iteration ≤ 1` whatever `refactorize` is. So the `QuasiNewton` case does not inherit a stale factorization either.

What follows is three documentation inaccuracies, one test gap, and two gaps that the PR's own argument opens but does not close.

---

### 1. The paragraph above the new one now contradicts it

`nonlinear_solver.jl:447`

> This is how the outcome of a solve is obtained **without holding on to a `NonlinearSolverState`**: the state is built here and handed to `status` afterwards.

That was true of both forms when it was written. It is now false of the form this PR adds — which is precisely the form that *does* hold a state. It needs to be scoped to the first and third forms.

### 2. The new paragraph sends the reader to an unexported name, and the `!!! info` block does not warn them

The added text promises that "a later `status(s, state)` returns what this returned". The `!!! info` block below it exists to name the unexported things a reader will reach for — it lists `isconverged` and `isstalled` — but not `status`, even though the CHANGELOG's own case for this PR is that `status(s, state)` "is unexported, so the pattern reaches past the public surface". A reader who follows the new sentence gets an `UndefVarError` and no hint from the block written to prevent exactly that. `status` belongs in that list.

### 3. Unused import in the first doctest's setup

`nonlinear_solver.jl:468` — the first jldoctest's setup gained `, status`, but that block only calls `isconverged`. The second block, which does use it, has its own setup. Revert the first.

### 4. The four-argument form is untested

Every new assertion (`nonlinear_solver_tests.jl:331-343`) uses the three-argument form. So forwarding `params` past the state — which, besides the two calls it wraps, is the *only* thing the new method does — is not covered. A dropped `params` would pass the whole new testset.

The natural home is the existing `"The solve! wrapper forwards params, a state and solver keywords"` testset, which already does this for `solve!` and whose `Fp!` discriminates: a dropped `params` arrives as `NullParameters`, which has no `.a`.

### 5. `solve_with_status!` still does not mirror `solve!` at the `prob` form

This is the one hole left in the PR's own thesis. `solve!(x, prob, method, args...)` forwards its positional arguments, and its docstring says so explicitly:

> The remaining positional arguments are passed on to `solve!(::AbstractArray, ::NonlinearSolver, ::NonlinearSolverState, params)`, so they are either nothing at all, the `params` of the problem, or a `NonlinearSolverState` followed by `params`.

`solve_with_status!(x, prob, method, params; kwargs...)` takes only `params`, so a caller who carries that sentence across to `solve_with_status!` — which is exactly what this PR invites them to do — gets a `MethodError`. One-line fix (`params=NullParameters()` → `args...`); verified it introduces no ambiguity against either solver-taking method.

### 6. The status is built twice per solve

Not introduced here, but this is the path the PR added to cut per-solve overhead, so it is worth closing at the same time. `solve!` builds a `NonlinearSolverStatus` at `nonlinear_solver.jl:311` to hand to `nonlinear_solver_warnings` and `print_status`, then discards it; `solve_with_status!` then asks `status(s, state)` for an identical one. That is five `l2norm` passes per solve spent reproducing a value that had just been thrown away — a larger per-solve cost than the state allocation this PR exists to remove.

Both can share one body (`_solve_core!`, following the `_bracket_core`/`_triple_point_core` idiom already in `src/bracketing/`) that returns the status the loop built: `solve!` returns `x`, `solve_with_status!` returns the status. The returned value is unchanged, since nothing touches the state between the end of the loop and the caller — and the PR's own `@test status(sv, statev) == stv` is what keeps that true, which makes that assertion more valuable after the change than before.

---

### One thing to resolve

The CHANGELOG says "22 of the 33 `NonlinearSolver` call sites in GeometricIntegrators **and GeometricIntegratorsBase**"; the PR description says "22 of the 33 `NonlinearSolver` call sites" in GeometricIntegrators alone. One of the two is wrong. I cannot see the downstream repositories from here, so I have left both as they are.

### Verification of the fixes

Fixes for all six are prepared. With them: the full suite passes (3062 assertions in *Nonlinear Solvers*, +5), `Documenter.doctest(SimpleSolvers)` passes, the full docs build produces no new warnings, and `Test.detect_ambiguities(SimpleSolvers)` is empty. No timings or tolerances were re-measured.

*Version `0.12.1` is right for this under the Julia 0.x convention; nothing there needs changing.*

michakraus:
Resolving the count discrepancy the review flagged, since it needed the downstream repositories to settle. **The CHANGELOG is right and the PR description was wrong**: the 33 spans both repositories, not GeometricIntegrators alone.

Counted in the migrated trees:

| | state-taking (4-arg) | state-building (3-arg) | dead code | total |
|---|---|---|---|---|
| GeometricIntegratorsBase | 7 | 0 | 0 | 7 |
| GeometricIntegrators | 15 | 4 | 7 | 26 |
| **total** | **22** | **4** | **7** | **33** |

So `22 of the 33 … in GeometricIntegrators and GeometricIntegratorsBase` is exact. Two caveats on that 33 worth stating, since the number is now in a released CHANGELOG:

- **7 of the 33 are dead code** — `src/integrators/vprk/`, reachable only through `src/integrators/VPRK.jl`, whose every `include` is commented out and which is itself never included. They were left on `solve!` deliberately, because converting them would imply they are live. So the honest form is *22 of the 26 live call sites*, which is a stronger number than the one shipped, not a weaker one.
- The 4 state-building sites are not oversights: DIRK's per-stage solvers get a `NullSolverState` from their `SingleStageSolvers` wrapper, and a `ProjectionIntegrator` has no `solverstate` field at all. Neither has a state to reuse.

No edit proposed — the shipped sentence is true as written, and 0.12.1 is already going out.

Two notes back on the merged fixes, from the downstream side:

**Finding 6 made the downstream migration free.** Before it, converting a call site from `solve!` to `solve_with_status!` bought the status at the cost of a second `NonlinearSolverStatus` construction per solve — which in a time-stepping loop is per step, and larger than the state allocation the PR set out to remove. With `_solve_core!` returning the status the loop already built, that cost is gone and the conversion is genuinely zero-overhead. It is the finding with the most downstream effect of the six, and it was not one of the review's four original items.

**The PGLRK case is the one place `status(s, state)` is still called by hand downstream**, and finding 6 is what makes it safe. PGLRK's stage solves happen inside a merit function that `SimpleSolvers.bisection` calls once per trial λ, so the *probe* statuses must not reach the integrator's failure hook — only the accepted λ's does. That is read off the persistent state after the fact, which is exactly the "nothing touches the `state` between the end of the loop and the caller" property `_solve_core!`'s docstring now asserts and the test suite pins.

Verified downstream after the merge: GeometricIntegratorsBase, GeometricIntegrators and NonlinearIntegrators all pass; GeometricIntegrators reports identical pass and broken counts across all 28 testsets against a pre-change worktree; and `LotkaVolterra2d` over 1000 steps is bit-identical for `Gauss(1)`, `Gauss(2)`, `Gauss(3)`, `ImplicitMidpoint` and `SRK3`, compared as raw `Float64` bit patterns.

